PR #13, opened 2021-07-13
Added Ingress Nginx Controller module and set default values for all …

UlaganathanNamachivayam:
…helm chart variables in root variables.tf file. Fixed naming convention limit error for public node group

*Description of changes:*

  1) Nginx ingress controller helm module is added.
  2) Fixed public node group name limit error.
  3) Set default values for helm chart variables in sources/variables.tf file.


source/kubeconfig_aws001-preprod-dev-eks
```diff
@@ -4,8 +4,8 @@ kind: Config
...
+    server: https://E2A6FA7A4933FCCBF024434B6B83F8C8.gr7.eu-west-1.eks.amazonaws.com
```
vara-bonthu, line 7: @UlaganathanNamachivayam  Could you please remove this file from this PR? We don't want to save this file

vara-bonthu (changes requested):
source/kubeconfig_aws001-preprod-dev-eks Please delete this file from this PR. 
 PR LGTM apart form the above issue

PR #22, opened 2021-09-02
Add metadata options to launch template

kpena027:
*Description of changes:*
Enables the ability to specify metadata options in launch template. Passes default options into source/main.tf as an example and sets variables to have default values. Ability to pass in metadata is useful for tightening security when using IRSA, see https://docs.aws.amazon.com/eks/latest/userguide/best-practices-security.html

### DIFF
--- a/modules/launch-templates/main.tf
+++ b/modules/launch-templates/main.tf
@@ -66,6 +66,12 @@ resource "aws_launch_template" "default" {
     enabled = true
   }
 
+  metadata_options {
+    http_endpoint               = var.http_endpoint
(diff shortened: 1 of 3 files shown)


vara-bonthu:
@kenerwin88 Thanks for your contribution. I see that you work for AWS. Could you please ping me on Slack (@vabonthu) so that i can introduce you to internal EKS accelerator team. We are currently working on refactoring the whole framework to make it more dynamic and it's still in progress. You can look at the feature branch in this repo which will give you some idea.

PR #23, opened 2021-09-02
porting vpc module to aws-ia vpc module

jaymccon:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/vpc.tf
+++ b/vpc.tf
@@ -33,63 +33,64 @@ module "vpc-label" {
 # VPC, SUBNETS AND ENDPOINTS DEPLOYED FOR FULLY PRIVATE EKS CLUSTERS
 # ---------------------------------------------------------------------------------------------------------------------
 module "vpc" {
-  create_vpc = var.create_vpc
-  source     = "terraform-aws-modules/vpc/aws"
-  version    = "v3.2.0"
-  name       = module.vpc-label.id
-  cidr       = var.vpc_cidr_block
-  azs        = data.aws_availability_zones.available.names
+  region               = data.aws_region.current.name
+  create_vpc           = var.create_vpc
+  source               = "aws-ia/vpc/aws"
+  version              = "0.1.1"
+  name                 = module.vpc-label.id
+  cidr                 = var.vpc_cidr_block
+  availability_zones   = data.aws_availability_zones.available.names
   # Private Subnets
-  private_subnets     = var.enable_private_subnets ? var.private_subnets_cidr : []
+  private_subnets_a   = var.enable_private_subnets ? var.private_subnets_cidr : []
   private_subnet_tags = var.enable_private_subnets ? local.private_subnet_tags : {}
 
   # Public Subnets
-  public_subnets     = var.enable_public_subnets ? var.public_subnets_cidr : []
-  public_subnet_tags = var.enable_public_subnets ? local.public_subnet_tags : {}
-
-  enable_nat_gateway = var.enable_nat_gateway ? var.enable_nat_gateway : false
-  single_nat_gateway = var.single_nat_gateway ? var.single_nat_gateway : false
-  create_igw         = var.enable_public_subnets && var.create_igw ? var.create_igw : false
-
-  enable_vpn_gateway              = false
-  create_egress_only_igw          = false
-  create_database_subnet_group    = false
-  create_elasticache_subnet_group = false
-  create_redshift_subnet_group    = false
+  public_subnets       = var.enable_public_subnets ? var.public_subnets_cidr : []
+  public_subnet_tags   = var.enable_public_subnets ? local.public_subnet_tags : {}
 
   enable_dns_hostnames = true
   enable_dns_support   = true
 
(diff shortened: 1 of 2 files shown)


vpc.tf
```diff
@@ -33,63 +33,64 @@ module "vpc-label" {
...
-  create_elasticache_subnet_group = false
-  create_redshift_subnet_group    = false
+  public_subnets       = var.enable_public_subnets ? var.public_subnets_cidr : []
+  public_subnet_tags   = var.enable_public_subnets ? local.public_subnet_tags : {}
 
   enable_dns_hostnames = true
   enable_dns_support   = true
 
-  # Enabling Custom Domain name servers
+  // TODO: add support for these fields to the aws-ia/vpc module
```
jaymccon, line 54: from here down to line 93 I commented out things that are not available in the aws-ia vpc module. @vara-bonthu can I get your input on which of these features are required. At a glance I'm guessing the default security group, and anything with a conditional (like `enable_nat_gateway`) are required, are there any others ?

vpc.tf
```diff
@@ -33,63 +33,64 @@ module "vpc-label" {
...
-
-  enable_nat_gateway = var.enable_nat_gateway ? var.enable_nat_gateway : false
-  single_nat_gateway = var.single_nat_gateway ? var.single_nat_gateway : false
-  create_igw         = var.enable_public_subnets && var.create_igw ? var.create_igw : false
-
-  enable_vpn_gateway              = false
-  create_egress_only_igw          = false
-  create_database_subnet_group    = false
-  create_elasticache_subnet_group = false
-  create_redshift_subnet_group    = false
```
vara-bonthu, line 58: @jaymccon You are right - We don't need these commented lines.

vpc.tf
```diff
@@ -33,63 +33,64 @@ module "vpc-label" {
...
+  private_subnets_a   = var.enable_private_subnets ? var.private_subnets_cidr : []
   private_subnet_tags = var.enable_private_subnets ? local.private_subnet_tags : {}
 
   # Public Subnets
-  public_subnets     = var.enable_public_subnets ? var.public_subnets_cidr : []
-  public_subnet_tags = var.enable_public_subnets ? local.public_subnet_tags : {}
-
-  enable_nat_gateway = var.enable_nat_gateway ? var.enable_nat_gateway : false
-  single_nat_gateway = var.single_nat_gateway ? var.single_nat_gateway : false
-  create_igw         = var.enable_public_subnets && var.create_igw ? var.create_igw : false
```
vara-bonthu, line 52: @jaymccon I think we need a feature in our VPC module to create a NAT and Internet gateway. Customers might using proxy servers instead of NAT or Internet gateway. However, This is a switch on/off feature and its helpful for workshops and demos

vpc.tf
```diff
@@ -33,63 +33,64 @@ module "vpc-label" {
...
+  //  # Enabling Custom Domain name servers
   //  enable_dhcp_options              = true
   //  dhcp_options_domain_name         = "service.consul"
   //  dhcp_options_domain_name_servers = ["127.0.0.1", "10.10.0.2"]
-
-  # VPC Flow Logs (Cloudwatch log group and IAM role will be created)
-  enable_flow_log                      = false
-  create_flow_log_cloudwatch_log_group = false
-  create_flow_log_cloudwatch_iam_role  = false
-  flow_log_max_aggregation_interval    = 60
```
vara-bonthu, line 72: This feature may not be required at this stage but its really useful feature for our AWS-IA VPC module. Every customer would like to create a VPC flow logs

vpc.tf
```diff
@@ -33,63 +33,64 @@ module "vpc-label" {
...
-      to_port     = 0
-      cidr_blocks = var.vpc_cidr_block
-  }]
-  default_security_group_egress = [
-    {
-      from_port   = 0
-      to_port     = 0
-      protocol    = -1
-      cidr_blocks = "0.0.0.0/0"
-  }]
```
vara-bonthu, line 92: @jaymccon Currently we have hardcoded the VPC endpoint creation. We we build fully private clusters without public subnets then we need loads of VPC Endpoints for various AWS services. 
Customer might create these endpoints as a part of VPC creation since its belong to specific VPC. 

The above code is creating a new sec group for all VPC endpoints and allowing an inbound from VPC CIDR. This is very much needed when we want to make use of VPC endpoints in Private subnets

vpc.tf
```diff
@@ -33,63 +33,64 @@ module "vpc-label" {
...
 module "vpc" {
-  create_vpc = var.create_vpc
-  source     = "terraform-aws-modules/vpc/aws"
-  version    = "v3.2.0"
-  name       = module.vpc-label.id
-  cidr       = var.vpc_cidr_block
-  azs        = data.aws_availability_zones.available.names
+  region               = data.aws_region.current.name
+  create_vpc           = var.create_vpc
+  source               = "aws-ia/vpc/aws"
```
antonbabenko, line 38: Hmm, do you really want to make `just another Terraform AWS module for VPC`?

PR #66, opened 2021-11-04
Adding support for ArgoCD bootstrap apps.

kcoleman731:
*Issue #, if available:*

*Description of changes:*

This pull request adds support for bootstrapping an EKS cluster with multiple workloads via an ArgoCD App of Apps repo. 


kubernetes-addons/argocd/main.tf
```diff
@@ -73,5 +73,42 @@ resource "helm_release" "argocd" {
       value = each_item.value.value
     }
   }
+}
+
+# ---------------------------------------------------------------------------------------------------------------------
+# ArgoCD App of Apps Bootstrapping
+# ---------------------------------------------------------------------------------------------------------------------
 
+resource "kubernetes_manifest" "argocd_application" {
```
kcoleman731, line 82: @vara-bonthu lets discuss

variables.tf
```diff
@@ -371,3 +371,8 @@ variable "argocd_helm_chart" {
   default     = {}
   description = "ARGO CD Kubernetes Addon Configuration"
 }
+variable "argocd_applications" {
+  type    = any
```
askulkarni2, line 375: Should we be a little prescriptive on the input by changing the type to `map`?

kubernetes-addons/argocd/variables.tf
```diff
@@ -20,3 +20,8 @@ variable "argocd_helm_chart" {
   type    = any
   default = {}
 }
+
+variable "argocd_applications" {
+  type    = any
```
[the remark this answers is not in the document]
vara-bonthu: `map` wont work here. We are passing map of dynamic objects so `any` is the only option.

kubernetes-addons/argocd/main.tf
vara-bonthu: Just seen the limitation on this from Hasicorp blog. We will find a workaround. 
`Access to the Kubernetes API is required at plan time` . This means that a Kubernetes cluster must be created and available in a separate Terraform apply operation before you can apply a kubernetes_manifest resource. This is because the provider has to fetch the OpenAPI specification from the Kubernetes API to generate the type information required by Terraform.

vara-bonthu:
LGTM

vara-bonthu (approved):
LGTM

PR #68, opened 2021-11-11
Added KEDA helm addon for Kubernetes-based Event Driven Autoscaler

vara-bonthu:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/kubernetes-addons/irsa/main.tf
+++ b/kubernetes-addons/irsa/main.tf
@@ -17,21 +17,21 @@
  */
 
 # Kubernetes Namesapce
-resource "kubernetes_namespace" "ns" {
+resource "kubernetes_namespace" "add_on_ns" {
   metadata {
     name = var.kubernetes_namespace
 
     labels = {
-      "app.kubernetes.io/managed-by" = "terraform-aws-eks-accelerator"
+      "app.kubernetes.io/managed-by" = "terraform-ssp-amazon-eks"
     }
   }
 }
 
 # Kubernetes service account
-resource "kubernetes_service_account" "keda_sa" {
+resource "kubernetes_service_account" "add_on_sa" {
   metadata {
     name        = var.kubernetes_service_account
-    namespace   = kubernetes_namespace.ns.id
+    namespace   = kubernetes_namespace.add_on_ns.id
     annotations = { "eks.amazonaws.com/role-arn" : aws_iam_role.irsa.arn }
   }
   automount_service_account_token = true
--- a/kubernetes-addons/irsa/main.tf
+++ b/kubernetes-addons/irsa/main.tf
@@ -1,0 +1,54 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: MIT-0
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this
+ * software and associated documentation files (the "Software"), to deal in the Software
+ * without restriction, including without limitation the rights to use, copy, modify,
+ * merge, publish, distribute, sublicense, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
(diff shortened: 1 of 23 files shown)


kubernetes-addons/keda/keda-values.yaml
```diff
@@ -0,0 +1,25 @@
+
+image:
```
kcoleman731, line 2: If we have overrides, let's use the values.yaml file. Otherwise no reason to put default values in here.

kubernetes-addons/irsa/main.tf
```diff
@@ -0,0 +1,54 @@
...
+    name = var.kubernetes_namespace
+
+    labels = {
+      "app.kubernetes.io/managed-by" = "terraform-aws-eks-accelerator"
+    }
+  }
+}
+
+# Kubernetes service account
+resource "kubernetes_service_account" "keda_sa" {
```
kcoleman731, line 31: keda_sa -> sa or add_on_sa

kubernetes-addons/irsa/variables.tf
```diff
@@ -0,0 +1,46 @@
...
+
+variable "kubernetes_namespace" {
+  description = "Kubernetes Namespace name"
+}
+
+variable "kubernetes_service_account" {
+  description = "Kubernetes Service Account Name"
+}
+
+variable "eks_cluster_name" {
```
kcoleman731, line 27: cluster_name or cluste_id ... btw we have inconsistency in a bunch of places.

kubernetes-addons/keda/locals.tf
```diff
@@ -0,0 +1,79 @@
...
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+locals {
+  keda_service_account_name = "keda-operator"
+  keda_namespace            = "keda-ns"
+
+  irsa_set_values = [{
```
kcoleman731, line 23: No need to use set. Just use `default_keda_helm_values`. You are setting the same values in different places.

kubernetes-addons/keda/main.tf
```diff
@@ -0,0 +1,144 @@
...
+
+resource "aws_iam_policy" "keda_irsa" {
+  count = var.keda_create_irsa ? 1 : 0
+
+  name        = "${var.eks_cluster_name}-${local.keda_helm_app["name"]}-irsa"
+  path        = var.iam_role_path
+  description = "KEDA IAM role policy for SQS and CloudWatch"
+
+  policy = <<EOF
+{
```
kcoleman731, line 101: can we put these in json files in the module?

kcoleman731 (commented):
Few comments for ya!

kubernetes-addons/irsa/main.tf
vara-bonthu: good spot. updated

kubernetes-addons/irsa/variables.tf
vara-bonthu: eks_cluster_name is used in many places. May be we can change this `eks_cluster_id` as a part of another PR

kubernetes-addons/keda/locals.tf
vara-bonthu: Ok. I thought about it a lot here and we need `set` here to ensure our `serviceAccount.create` and `serviceAccount.name` when `var.keda_create_irsa = true` 

`default_keda_helm_values` will be overwritten if customer sends his own values.yaml and we will loose control.

I am using set but i am also validating this before applying in `main.tf` line61. With this approach we can control better

kubernetes-addons/keda/main.tf
vara-bonthu: done

kubernetes-addons/irsa/variables.tf
kcoleman731: Agree.  Let's do it.

kcoleman731 (approved):

PR #71, opened 2021-11-13
Feature/84 audit add on values

kcoleman731:
*Issue #, if available:*

*Description of changes:*

This pull request brings consistency to the way we leverage values in our helm chart modules. Previously, values resided in multiple places, including yaml files or the `set` operator.

We also had a fair amount of default helm values that had been copied and pasted into yaml files. This PR removes all default values and only keeps values which change default behavior. 

Lastly, the pr also renames alb-ingress-controller add-on to aws-load-balancer-controller to reflect the current name of the controller. 


vara-bonthu:
LGTM 👍

vara-bonthu (approved):
LGTM

PR #72, opened 2021-11-17
Feature/gitops add on configuration

kcoleman731:
*Issue #, if available:*

*Description of changes:*

This PR proposes changes needed to enable support for managing add-ons via GitOps with ArgoCD. From a top level module perspective, it introduces a new property `argocd_manage_add_ons`. This property indicates that add-ons should be managed via ArgoCD and that Helm charts for each enabled add-on should not be directly applied via Terraform. 

Additionally, it adds a `argocd_gitops_config` output property to the `aws-for-fluentbit`,  add-on (other add-ons will be supported in subsequent PRs). This property is used to build the values map that is the passed to the ArgoCD root Application resource. 

Lastly, customers may indicate which ArgoCD application serves as their root add-on Application via an `add_on_application` flag. ArgoCD configuration will now look like the following:

```
argocd_enable           = true
argocd_manage_add_ons   = true
argocd_applications     = {
  infra = {
    namespace             = "argocd"
    path                  = "<path>"
    repo_url              = "<repo_url>"
    target_revision       = "HEAD"
    destination           = "https://kubernetes.default.svc"
    project               = "default"
    values                = {}
    add_on_application    = true # Indicates the root add-on application. 
  }
}
```


docs/add-ons/argocd.md
```diff
@@ -49,3 +49,30 @@ argocd_applications = {
...
+    path                  = "<path>"
+    repo_url              = "<repo_url>"
+    target_revision       = "HEAD"
+    destination           = "https://kubernetes.default.svc"
+    project               = "default"
+    values                = {}
+    add_on_application    = true # Indicates the root add-on application. 
+  }
+}
+```
```
askulkarni2, line 78: Maybe just call out that alternatively if customers wish to do a CD pipeline (like GitLab) that's possible by using the default behavior.

askulkarni2 (commented):
looking good so far...

kubernetes-addons/aws-for-fluentbit/locals.tf
```diff
@@ -15,8 +16,6 @@ locals {
     namespace                                 = "kube-system"
```
vara-bonthu, line 16: We need to move Addons to its dedicated namespaces. Majority of them are suing `kube-system`. We can pick this up a seprate PR

vara-bonthu (approved):
LGTM

docs/add-ons/argocd.md
kcoleman731: Agree - although I plan to add this content to a top level `add-ons` doc that discusses the different approaches.

kubernetes-addons/aws-for-fluentbit/locals.tf
kcoleman731: agree.

PR #74, opened 2021-11-20
Adding GitOps config for additional add-ons

kcoleman731:
*Issue #, if available:*

*Description of changes:*

Adds functionality for managing add-on via GitOps for:

- [x] agones 
- [x] aws-for-fluent-bit
- [x] aws-load-balancer-controller
- [x] cert-manager
- [x] cluster-autoscaler


locals.tf
```diff
@@ -84,8 +84,12 @@ locals {
...
-  # Configuration for managing add-ons via GitOps.
-  gitops_add_on_config = {
-    awsForFluentBit = var.aws_for_fluentbit_enable ? module.aws_for_fluent_bit[0].gitops_config : null
+  # Configuration for managing add-ons via ArgoCD.
+  argocd_add_on_config = {
+    agones                    = var.agones_enable ? module.agones[0].argocd_gitops_config : {}
+    awsForFluentBit           = var.aws_for_fluentbit_enable ? module.aws_for_fluent_bit[0].argocd_gitops_config : null
+    awsLoadBalancerController = var.aws_lb_ingress_controller_enable ? module.aws_load_balancer_controller[0].argocd_gitops_config : null
+    certManager               = var.cert_manager_enable ? module.cert_manager[0].argocd_gitops_config : null
+    clusterAutoscaler         = var.cluster_autoscaler_enable ? module.cluster_autoscaler[0].argocd_gitops_config : null
```
vara-bonthu, line 93: Should we make these variables naming convention as per terraform for consistency?  small case with underscores
`awsForFluentBit` to `aws_for_fluentbit`

vara-bonthu (commented):
I added one comment for variables apart from that everything looks good to me.

locals.tf
kcoleman731: my thinking is that this object gets passed into Argo via the values map. So the variable consistency should likely align with K8s yaml? I.E. camel case?

locals.tf
```diff
@@ -84,8 +84,12 @@ locals {
   service_account_amp_ingest_name = format("%s-%s", module.aws_eks.cluster_id, "amp-ingest")
   service_account_amp_query_name  = format("%s-%s", module.aws_eks.cluster_id, "amp-query")
 
-  # Configuration for managing add-ons via GitOps.
-  gitops_add_on_config = {
-    awsForFluentBit = var.aws_for_fluentbit_enable ? module.aws_for_fluent_bit[0].gitops_config : null
+  # Configuration for managing add-ons via ArgoCD.
+  argocd_add_on_config = {
+    agones                    = var.agones_enable ? module.agones[0].argocd_gitops_config : {}
```
askulkarni2, line 89: Shouldn't this be `var.agones_enable ? module.agones[0].argocd_gitops_config : null` like the others?

askulkarni2 (changes requested):
Looks good. One minor comment.

Can you update the top-level readme as well and a GitOps section?

locals.tf
vara-bonthu: Got it! 
We can deal with this when we create a Helm module for this

locals.tf
kcoleman731: good catch

kcoleman731:
> 

Agree. Need to expand our add-ons guide in docs. But in subsequent docs pr.

askulkarni2 (approved):
:shipit:

locals.tf
```diff
@@ -84,8 +84,12 @@ locals {
...
-  # Configuration for managing add-ons via GitOps.
-  gitops_add_on_config = {
-    awsForFluentBit = var.aws_for_fluentbit_enable ? module.aws_for_fluent_bit[0].gitops_config : null
+  # Configuration for managing add-ons via ArgoCD.
+  argocd_add_on_config = {
+    agones                    = var.agones_enable ? module.agones[0].argocd_gitops_config : null
+    awsForFluentBit           = var.aws_for_fluentbit_enable ? module.aws_for_fluent_bit[0].argocd_gitops_config : null
+    awsLoadBalancerController = var.aws_lb_ingress_controller_enable ? module.aws_load_balancer_controller[0].argocd_gitops_config : null
+    certManager               = var.cert_manager_enable ? module.cert_manager[0].argocd_gitops_config : null
+    clusterAutoscaler         = var.cluster_autoscaler_enable ? module.cluster_autoscaler[0].argocd_gitops_config : null
```
vara-bonthu, line 93: I think we also need to another condition here  `var.argocd_manage_add_ons`

e.g., `agones                    = var.agones_enable && var.argocd_manage_add_ons ? module.agones[0].argocd_gitops_config : null`

vara-bonthu (changes requested):
I just added another comment. I found this issue during my tests

locals.tf
kcoleman731: If `argocd_manage_add_ons` is set to `false` (default), this value is not used...so probably not needed?

PR #139, opened 2021-12-22
tfdocs updates for only feature branches; tf plan to write to PR

vara-bonthu:
*Issue #, if available:*

*Description of changes:*
- tfdocs workflow runs only when the PR is raised from the same repo using feature branches (not from the forked repo)

### DIFF
--- a/.github/workflows/terraform-docs.yaml
+++ b/.github/workflows/terraform-docs.yaml
@@ -2,20 +2,33 @@ name: terraform-docs
 on:
   push:
     branches: [ main ]
+  pull_request:
+    types: [assigned, opened, synchronize, reopened, edited]
 
 jobs:
   docs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        if: github.event_name == 'push'
         with:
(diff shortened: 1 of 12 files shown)


.github/workflows/terraform-docs.yaml
```diff
@@ -2,20 +2,33 @@ name: terraform-docs
 on:
   push:
     branches: [ main ]
+  pull_request:
```
kcoleman731, line 5: don't think we want this.

kcoleman731 (commented):
Don't think we want this. Why not just have docs update on merge to main?

.github/workflows/terraform-docs.yaml
vara-bonthu: Github workflow unable to push commits to MAIN and the workflow is failing due to our restrictions.
- No commits to main
- at-least one reviewer is required

It's also good to see what's being pushed to MAIN by looking at the PR. Hence i changed this to  on pull request but this will be executed only when the PR raised within the same repo (tfdocs wont be run on PRs raised from fork)

This PR worked with the latest workflow

vara-bonthu:
Add this change to k8s addon refactor PR hence closed

PR #164, opened 2021-12-29
E2e test update

vara-bonthu:
*Issue #, if available:*

*Description of changes:*
- e2e test updates with new add-on approach


test/e2e/vpc/main.tf
```diff
@@ -16,33 +16,24 @@
...
 }
 
 terraform {
   backend "s3" {
-    bucket = "terraform-ssp-github-actions-state"
-    key    = "e2e/vpc/terraform-main.tfstate"
-    region = "us-west-2"
+    bucket = ""
+    key    = ""
+    region = ""
```
askulkarni2, line 27: Is this required?

vara-bonthu: Yes, unfortunately we need those values to be empty otherwise it will throw an error

kcoleman731: We sure about this? This doesn't seem right....can we just use variables here with defaults?

vara-bonthu: @askulkarni2 Correction here. You are right that we may not need this anymore. I just updated the PR now.

Zvikan (approved):

PR #167, opened 2021-12-30
Adding explicit namespace to prometheus

kcoleman731:
*Issue #, if available:*

*Description of changes:*

Fixes the Prometheus destroy issue. 

### DIFF
--- a/modules/aws-managed-prometheus/locals.tf
+++ b/modules/aws-managed-prometheus/locals.tf
@@ -4,7 +4,7 @@ locals {
   irsa_config = {
     ingest = {
       service_account             = "amp-ingest",
-      create_kubernetes_namespace = true,
+      create_kubernetes_namespace = false,
       irsa_iam_policies           = [aws_iam_policy.ingest.arn]
 
     },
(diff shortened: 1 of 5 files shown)


vara-bonthu (commented):
I am happy with this change if this has been tested with Amazon_Prometheus enabled and disabled with Prometheus server

vara-bonthu (approved):

modules/aws-managed-prometheus/locals.tf
```diff
@@ -4,7 +4,7 @@ locals {
   irsa_config = {
     ingest = {
       service_account             = "amp-ingest",
-      create_kubernetes_namespace = true,
+      create_kubernetes_namespace = false,
```
vara-bonthu, line 7: Amazon Managed Prometheus module creation is failing due to prometheus namespace not found.

│ Error: namespaces "prometheus" not found
│
│ with module.aws-eks-accelerator-for-terraform.module.aws_managed_prometheus[0].module.irsa["query"].kubernetes_service_account_v1.irsa[0],
│ on /Users/vabonthu/Documents/GITHUB/aws-eks-accelerator-for-terraform/modules/irsa/main.tf line 30, in resource "kubernetes_service_account_v1" "irsa":
│ 30: resource "kubernetes_service_account_v1" "irsa" {
│
╵

vara-bonthu (changes requested):
I think we may need a different approach to solve both AMP and Prometheus. 

We can move Amazon Managed Prometheus to Prometheus K8s module. Create dependencies between the modules. However we may need to document the dependency between these modules

AMP can also be used without Prometheus by leveraging OTEL. We may need a different solution for that.

PR #169, opened 2022-01-04
Analytics examples

vara-bonthu:
*Issue #, if available:*

*Description of changes:*

This feature allows users to use the existing Amazon prometheus module to create Workspaces. Users can pass existing workspace id to Prometheus Server or it will create new one by Prometheus Add-on module if none provided

- Moved AMP resources to Prometheus module
- Updated Analytics examples
- Updated docs


docs/add-ons/prometheus.md
```diff
@@ -8,20 +8,41 @@ This project provides support for installing a open source Prometheus server in
...
+```hcl-terraform
+  #---------------------------------------
+  # Prometheus Server integration with Amazon Prometheus
+  #---------------------------------------
+  enable_amp_for_prometheus = true
+  # amazon_prometheus_workspace_id = module.aws-eks-accelerator-for-terraform.amazon_prometheus_workspace_id # Optional variable to use existing AMP workspace or it will create a new one
+  enable_prometheus = true
+  # Optional Map value
+  prometheus_helm_config = {
+    name       = "prometheus"                                         # (Required) Release name.
```
kcoleman731, line 30: Do we need to advanced config here?

modules/kubernetes-addons/prometheus/locals.tf
```diff
@@ -72,10 +47,50 @@ locals {
...
+    {
+      name  = "server.remoteWrite[0].url"
+      value = local.amp_workspace_url
+    },
+    {
+      name  = "server.remoteWrite[0].sigv4.region"
+      value = data.aws_region.current.id
+  }] : []
+
+  irsa_config = {
```
kcoleman731, line 76: IMHO this config is necessary.  Just makes the code more confusing to read.  Just create two Irsa module blocks.

modules/kubernetes-addons/main.tf
```diff
@@ -152,15 +152,15 @@ module "metrics_server" {
...
+  count          = var.enable_prometheus ? 1 : 0
+  source         = "./prometheus"
+  eks_cluster_id = var.eks_cluster_id
+  helm_config    = var.prometheus_helm_config
   #AWS Managed Prometheus Workspace
-  amazon_prometheus_workspace_id           = var.amazon_prometheus_workspace_id
-  amazon_prometheus_ingest_iam_role_arn    = var.amazon_prometheus_ingest_iam_role_arn
-  amazon_prometheus_ingest_service_account = var.amazon_prometheus_ingest_service_account
-  manage_via_gitops                        = var.argocd_manage_add_ons
+  enable_amp_for_prometheus      = var.enable_amp_for_prometheus
```
kcoleman731, line 160: I don't think this variable is necessary. If `amazon_prometheus_workspace_id` is not null, then we can assume that `enable_amp_for_prometheus` would be true right? 

Also consider changing `amazon_prometheus_workspace_id` to `amp_workspace_id`

modules/kubernetes-addons/prometheus/locals.tf
```diff
@@ -72,10 +47,50 @@ locals {
     operating_system = "linux",
   })]
 
+  amazon_prometheus_workspace_id           = var.amazon_prometheus_workspace_id == null && var.enable_amp_for_prometheus ? aws_prometheus_workspace.amp_workspace[0].id : var.amazon_prometheus_workspace_id
+  amp_workspace_url                        = var.enable_amp_for_prometheus && local.amazon_prometheus_workspace_id != null ? "https://aps-workspaces.${data.aws_region.current.id}.amazonaws.com/workspaces/${local.amazon_prometheus_workspace_id}/api/v1/remote_write" : null
+  amazon_prometheus_ingest_iam_role_arn    = var.enable_amp_for_prometheus ? module.irsa.*.ingest.irsa_iam_role_arn[0] : null
+  amazon_prometheus_ingest_service_account = local.irsa_config.ingest.service_account
```
kcoleman731, line 53: naming consistency with these variables. Either go with `amp` as a prefix or `amazon_prometheus`. I am partial to `amp` for brevity.

modules/kubernetes-addons/prometheus/main.tf
```diff
@@ -75,4 +75,52 @@ resource "helm_release" "prometheus" {
...
+    name = local.helm_config["namespace"]
+    labels = {
+      "app.kubernetes.io/managed-by" = "terraform-ssp-amazon-eks"
+    }
+  }
+}
+
+# Amazon Managed Prometheus Resources
+
+resource "aws_prometheus_workspace" "amp_workspace" {
```
kcoleman731, line 93: I don't think we want to provide customers with multiple places to create an amp_workspace in the framework. Either we keep it in the dedicated submodule or we move it here.

modules/kubernetes-addons/prometheus/main.tf
```diff
@@ -75,4 +75,52 @@ resource "helm_release" "prometheus" {
...
+  count = var.amazon_prometheus_workspace_id == null && var.enable_amp_for_prometheus ? 1 : 0
+  alias = format("%s-%s", "amp", var.eks_cluster_id)
+}
+
+module "irsa" {
+  for_each = { for k, v in local.irsa_config : k => v if var.enable_amp_for_prometheus }
```
kcoleman731, line 99: We can get rid of the for_each if we don't rely on the locals and just have two irsa blocks ;)

docs/add-ons/prometheus.md
vara-bonthu: I thought our plan is to update the docs with basic usage and advanced usage of each add-on? Is there any place where we can document this?

modules/kubernetes-addons/main.tf
vara-bonthu: We have three scenarios here...

1. Users passing existing AMP workspace ID to integrate with Prometheus (This will integrate existing AMP WS with Prometheus.)
`enable_prometheus=true` &&`enable_amp_for_prometheus = true` && `amazon_prometheus_workspace_id=something` 
2. Users enabling Prometheus only (This should ignore AMP config)
`enable_prometheus=true` && `enable_amp_for_prometheus = false`(default)
3.  Users don't pass AMP workspace but they enable AMP with Prometheus (This will create AMP and integrate with Prometheus. amazon_prometheus_workspace_id is null here)
    `enable_amp_for_prometheus = true` && `enable_prometheus=true` && `amazon_prometheus_workspace_id=null` 

Overall `enable_amp_for_prometheus = true` is critical to integrate the AMP workspace ID with Prometheus or not.
`amazon_prometheus_workspace_id` is used  to check whether to create a new AMP Woekspace or not.

modules/kubernetes-addons/prometheus/locals.tf
vara-bonthu: Sure, I have told by the teams not to use `amp` . I will change this to `amazon_prometheus`

modules/kubernetes-addons/prometheus/locals.tf
vara-bonthu: I thought first it may not be nice to duplicate IRSA module for each service account. I thought this is more cleaner.

modules/kubernetes-addons/prometheus/main.tf
vara-bonthu: As you know `kubernetes-addons` is an independent sub module. It can be used without any dependency on core module. Having this feature allows customer to create workspace if they don't provide any existing workspace id.

We can remove the core AMP module if that's not necessary

modules/kubernetes-addons/prometheus/main.tf
vara-bonthu: I can change this to two IRSA blocks

modules/kubernetes-addons/prometheus/locals.tf
vara-bonthu: removed and duplicated the irsa modules now

modules/kubernetes-addons/prometheus/main.tf
vara-bonthu: done

docs/add-ons/prometheus.md
kcoleman731: Fine by me

examples/advanced/k8s_addons/main.tf
```diff
@@ -308,8 +308,11 @@ module "kubernetes-addons" {
   #---------------------------------------
   # COMMUNITY PROMETHEUS ENABLE
   #---------------------------------------
-  enable_prometheus = true
+  # Amazon Prometheus Configuration to integrate with Prometheus Server Add-on
+  enable_amp_for_prometheus      = true
```
kcoleman731, line 312: don't think you need this property right? It is implied by the presence of the workspace_id?

modules/kubernetes-addons/prometheus/locals.tf
```diff
@@ -72,10 +47,36 @@ locals {
     operating_system = "linux",
   })]
 
+  amazon_prometheus_workspace_id           = var.amazon_prometheus_workspace_id != null && var.enable_amp_for_prometheus ? var.amazon_prometheus_workspace_id : null
```
kcoleman731, line 50: Isn't var.amazon_prometheus_workspace_id != null a sufficient check?

modules/kubernetes-addons/prometheus/locals.tf
```diff
@@ -72,10 +47,36 @@ locals {
...
+  amazon_prometheus_workspace_url          = var.enable_amp_for_prometheus && var.amazon_prometheus_workspace_id != null ? "https://aps-workspaces.${data.aws_region.current.id}.amazonaws.com/workspaces/${local.amazon_prometheus_workspace_id}/api/v1/remote_write" : null
```
kcoleman731, line 51: Is this value the same as `amazon_prometheus_workspace_endpoint`? If so, should we be passing that value instead of the workspace_id?

modules/kubernetes-addons/variables.tf
```diff
@@ -96,6 +96,12 @@ variable "cluster_autoscaler_helm_config" {
 }
 
 #-----------Amazon Managed Service for Prometheus-------------
+variable "enable_amp_for_prometheus" {
```
kcoleman731, line 99: We don't want to use `amp` right? So probably `enable_amazon_prometheus` is a better prop name. Matches the naming convention of the exported values as well.

kcoleman731 (commented):
Looking good. Few comments.

modules/kubernetes-addons/prometheus/locals.tf
vara-bonthu: yes that should do

modules/kubernetes-addons/prometheus/locals.tf
vara-bonthu: Not sure. I will try to print out to see what comes out

modules/kubernetes-addons/variables.tf
vara-bonthu: will change that

askulkarni2 (approved):

kcoleman731 (approved):

PR #248, opened 2022-02-11
Karpenter example with default provisioner

vara-bonthu:
### What does this PR do?

- This example shows a pattern on how to deploy multiple Karpenter Cluster Autoscaler Provisioners

### Motivation

- Users feedback on Karpenter to manage provisioners using Terraform

### More

- [x] Yes, I have added a new example under [examples](https://github.com/aws-samples/aws-eks-accelerator-for-terraform/tree/main/examples) to support my PR
- [x] Yes, I have updated the [docs](https://github.com/aws-samples/aws-eks-accelerator-for-terraform/tree/main/docs) for this feature

### For Moderators

### DIFF
--- a/examples/karpenter/README.md
+++ b/examples/karpenter/README.md
@@ -91,6 +91,7 @@ terraform destroy
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.0.1 |
 | <a name="requirement_aws"></a> [aws](#requirement\_aws) | >= 3.66.0 |
 | <a name="requirement_helm"></a> [helm](#requirement\_helm) | >= 2.4.1 |
+| <a name="requirement_kubectl"></a> [kubectl](#requirement\_kubectl) | >= 1.13.1 |
 | <a name="requirement_kubernetes"></a> [kubernetes](#requirement\_kubernetes) | >= 2.6.1 |
 
 ## Providers
@@ -98,6 +99,7 @@ terraform destroy
 | Name | Version |
 |------|---------|
 | <a name="provider_aws"></a> [aws](#provider\_aws) | >= 3.66.0 |
+| <a name="provider_kubectl"></a> [kubectl](#provider\_kubectl) | >= 1.13.1 |
 
 ## Modules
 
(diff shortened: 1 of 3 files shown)


examples/karpenter/README.md
```diff
@@ -72,19 +73,11 @@ You should see one Self-managed node up and running
...
-    $ cd examples/eks-cluster-with-karpenter/provisioners
-    $ kubectl apply -f default_provisioner.yaml
-
-#### Step9: Run this sample `deployment` to verify the Autoscaling triggered by Karpenter
-
-    $ cd examples/eks-cluster-with-karpenter/provisioners
-    $ kubectl apply -f sample_deployment.yaml
+# How to Destroy
 
+NOTE: Make sure you delete all the deployments which clean up the nodes spun up by Karpenter Autoscaler
```
askulkarni2, line 78: This is an important thing to discuss. Terraform does not address the drift when using Karpenter.

examples/karpenter/main.tf
```diff
@@ -14,6 +14,10 @@ terraform {
       source  = "hashicorp/helm"
       version = ">= 2.4.1"
     }
+    kubectl = {
+      source  = "gavinbunney/kubectl"
+      version = ">= 1.13.1"
+    }
```
askulkarni2, line 20: Are we okay to use a 3P provider? Or should we consider using [tfk8s](https://github.com/jrhouston/tfk8s) which is Terraform [recommended](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/manifest)?

examples/karpenter/README.md
vara-bonthu: That's why added this note to the users.
We should be ok as Terraform only cares about the nodes created by Terraform Node groups. It will not have any drift issues with Karpenter nodes as these are not joining the existing node groups.

Karpenter Nodes can only be deleted when you delete the deployments.

examples/karpenter/main.tf
vara-bonthu: I think we agreed to proceed with `gavinbunney/kubectl` as this works better than the Hashiciorp Kubernetes manifest. We introduced this in our TEAMS feature. This can handle any kubernetes object including the CRDs.

Hashicorp Kubernetes manifest resources has few limitations which is causing an issues for our plan. It has limitation ons CRDs as well. check this out https://www.hashicorp.com/blog/beta-support-for-crds-in-the-terraform-provider-for-kubernetes. 

yes, for now we will be using kubectl_manifest until we have better version from Hashicorp

askulkarni2 (approved):
looks good!

PR #261, opened 2022-02-15
Add Permission Boundary Support to AWS Load Balancer Controller

bobdoah:
### What does this PR do?

Allow a permissions boundary to be applied to the IRSA IAM role. Additionally, support applying it to the role created by the AWS load balancer controller addon.  

### Motivation

We use a IAM role to provision our Terraform which has permissions boundary attached. That policy must be applied to all IAM roles created by the Terraform. It is used to prevent privilege escalation in keeping with https://aws.amazon.com/premiumsupport/knowledge-center/iam-permission-boundaries/
 

### More

### For Moderators

### Additional Notes

I would be happy to add the same functionality to the other addons, as they're adopted by us. I do not have the capacity to test them at the moment.

### DIFF
--- a/modules/kubernetes-addons/main.tf
+++ b/modules/kubernetes-addons/main.tf
@@ -75,7 +75,7 @@ module "argo_rollouts" {
   source                        = "./argo-rollouts"
   eks_cluster_id                = var.eks_cluster_id
   helm_config                   = var.argo_rollouts_helm_config
-  irsa_iam_permissions_boundary = var.irsa_iam_permissions_boundary
+  irsa_iam_permissions_boundary = var.argo_rollouts_irsa_permissions_boundary
   tags                          = var.tags
   manage_via_gitops             = var.argocd_manage_add_ons
 }
--- a/modules/kubernetes-addons/variables.tf
+++ b/modules/kubernetes-addons/variables.tf
@@ -442,6 +442,12 @@ variable "argo_rollouts_helm_config" {
   description = "Argo Rollouts Helm Chart config"
 }
 
+variable "argo_rollouts_irsa_permissions_boundary" {
+  type        = string
+  default     = ""
+  description = "IAM Policy ARN for IRSA IAM role permissions boundary"
+}
+
 variable "argo_rollouts_irsa_policies" {
   type        = list(string)
   default     = []


modules/kubernetes-addons/aws-load-balancer-controller/README.md
```diff
@@ -58,6 +58,7 @@ Here is the link to get the AWS ELB [service annotations](https://kubernetes-sig
 
 
 <!--- BEGIN_TF_DOCS --->
+<!--- END_TF_DOCS --->
```
vara-bonthu, line 61: See the above comment

modules/kubernetes-addons/aws-load-balancer-controller/README.md
```diff
@@ -116,5 +118,3 @@ No requirements.
 | <a name="output_argocd_gitops_config"></a> [argocd\_gitops\_config](#output\_argocd\_gitops\_config) | Configuration used for managing the add-on with ArgoCD |
 | <a name="output_ingress_name"></a> [ingress\_name](#output\_ingress\_name) | AWS LoadBalancer Controller Ingress Name |
 | <a name="output_ingress_namespace"></a> [ingress\_namespace](#output\_ingress\_namespace) | AWS LoadBalancer Controller Ingress Namespace |
-
-<!--- END_TF_DOCS --->
```
vara-bonthu, line 120: This shouldn't be removed

vara-bonthu (changes requested):
Hi @bobdoah Thanks for your first contribution on our repo 👍🏼 
I understand this is useful feature to control the IAM roles. 

PR Looks good to me. I have one comment on docs that needs addressing

kcoleman731:
@bobdoah thank you for the PR! 

To aid in adding support to other add-ons, it probably makes sense to add permission boundary support the [`helm-addon` submodule](https://github.com/aws-samples/aws-eks-accelerator-for-terraform/tree/main/modules/kubernetes-addons/helm-addon). LBC add-on doesn't leverage the `helm-addon` submodule yet, but our plan is to update all add-ons which deploy helm charts to leverage it in the near future. We could prioritize updating the LBC add-on if you need.

bobdoah:
Thanks @vara-bonthu, @kcoleman731 for the prompt response.

I have corrected the documentation markers, an oversight on my part. I had thought I had inserted the docs in the correct place.

@kcoleman731 I have added support to the `helm-addon` module as suggested. It didn't look like there were many modules using that, at the moment. It would be great if you could prioritise the LBC change. My intent was to revisit the other modules as I adopted them.

bobdoah:
Realised I need to add this to the modules using `helm-addon` or they will not work.

vara-bonthu (approved):
Nice Work on this PR 👍🏼  LGTM!

bobdoah:
Any chance of getting this merged? Is there something I need to do fix the pr-test action?

PR #278, opened 2022-02-22
change create eks default

askulkarni2:
### What does this PR do?

Sets `create_eks` to `true` by default and removes unnecessary use of the variable from examples.

### Motivation

Customers will need the control plane in most cases when using the framework. Very few customers are going to leverage an existing control plane. To that end the default behavior should be `create_eks = true` as is documented in our [README](./README.md).

### More

- [N/A] Yes, I have added a new example under [examples](https://github.com/aws-samples/aws-eks-accelerator-for-terraform/tree/main/examples) to support my PR
- [N/A] Yes, I have created another PR for add-ons under [add-ons](https://github.com/aws-samples/ssp-eks-add-ons) repo (if applicable)
- [X] Yes, I have updated the [docs](https://github.com/aws-samples/aws-eks-accelerator-for-terraform/tree/main/docs) for this feature

### For Moderators
- [X] E2E Test successfully complete before merge?


vara-bonthu (changes requested):
There are lot of other occurrences of `create_eks` which needs fixing and testing the module before we merge this

askulkarni2:
> There are lot of other occurrences of `create_eks` which needs fixing and testing the module before we merge this

Not sure I understand, I am only the default value in the top level `variables.tf`. I am not changing the behavior of the module. Can you an example of what else needs fixing?

vara-bonthu (approved):
Got it! Thanks

PR #313, opened 2022-03-04
Fix keda role path

bobdoah:
The missing IAM role path is causing a plan with this module to fail.
This value is accepted by the IRSA module, so reinstate it and pass to
the module via the IRSA configuration.

### What does this PR do?

PR #312 introduced a breaking change. Remove of the `iam_role_path` prevents `terraform plan` from succeeding. Instead it fails with:

```
│ Error: Reference to undeclared input variable
│ 
│   on .terraform/modules/kubernetes-addons/keda/main.tf line 15, in resource "aws_iam_policy" "keda_irsa":
│   15:   path        = var.iam_role_path
│ 
│ An input variable with the name "iam_role_path" has not been declared. This
│ variable can be declared with a variable "iam_role_path" {} block.
╵
```
This change reintroduces that configuration parameter. It appears that it is supported as a configuration option by the IRSA module. This change also populates the IRSA module configuration with the value supplied. This enables the IRSA and associated IAM resources to be created under a particular path.

There were another couple of issues, so I aligned the code with what appears to be working under the `kubernetes_addons/vpa` module.

### Motivation

Breakage is blocking my current deployments.

### More

### For Moderators

### DIFF
--- a/modules/kubernetes-addons/keda/locals.tf
+++ b/modules/kubernetes-addons/keda/locals.tf
@@ -33,7 +33,7 @@ locals {
   ]
 
   irsa_config = {
-    kubernetes_namespace              = local.name
+    kubernetes_namespace              = local.helm_config["namespace"]
     kubernetes_service_account        = local.service_account_name
     create_kubernetes_namespace       = true
     create_kubernetes_service_account = true
@@ -41,6 +41,7 @@ locals {
     irsa_iam_permissions_boundary     = var.irsa_permissions_boundary
     eks_cluster_id                    = var.addon_context.eks_cluster_id
     iam_role_path                     = var.iam_role_path
+    eks_cluster_id                    = var.addon_context.eks_cluster_id
   }
 
   argocd_gitops_config = {
--- a/modules/kubernetes-addons/keda/locals.tf
+++ b/modules/kubernetes-addons/keda/locals.tf
@@ -40,6 +40,7 @@ locals {
     irsa_iam_policies                 = concat([aws_iam_policy.keda_irsa[0].arn], var.irsa_policies)
     irsa_iam_permissions_boundary     = var.irsa_permissions_boundary
     eks_cluster_id                    = var.addon_context.eks_cluster_id
+    iam_role_path                     = var.iam_role_path
   }
 
   argocd_gitops_config = {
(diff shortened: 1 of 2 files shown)


vara-bonthu:
@bobdoah Thanks for spotting the issue. Part of the fix was merged by another PR. Could you please rebase the PR please

bobdoah:
Done. I should perhaps point out that I also tried to pin to version 3.4.1, but the kube-state-metrics (0.0.2 in that version) module has an issue. It includes the `helm-addon` module, but isn't pinned to a particular version. It isn't compatible with the new changes.

modules/kubernetes-addons/keda/locals.tf
```diff
@@ -33,13 +33,15 @@ locals {
...
   irsa_config = {
-    kubernetes_namespace              = local.name
+    kubernetes_namespace              = local.helm_config["namespace"]
     kubernetes_service_account        = local.service_account_name
     create_kubernetes_namespace       = true
     create_kubernetes_service_account = true
     irsa_iam_policies                 = concat([aws_iam_policy.keda_irsa[0].arn], var.irsa_policies)
     irsa_iam_permissions_boundary     = var.irsa_permissions_boundary
     eks_cluster_id                    = var.addon_context.eks_cluster_id
+    iam_role_path                     = var.iam_role_path
```
vara-bonthu, line 43: iam_role_path is not required for IRSA config

vara-bonthu (changes requested):
One change required and we are good

modules/kubernetes-addons/keda/locals.tf
bobdoah: Not required, but it is supported: https://github.com/aws-samples/aws-eks-accelerator-for-terraform/blob/9684d0a106d43719e3f50adcdf3d6e0d9b4a85ca/modules/irsa/variables.tf#L41

I guess I'm wondering, how I would change it, as there doesn't appear to be an override/merge for `local.irsa_config` . In addition, if I change it for the `"aws_iam_policy" "keda_irsa"` would I not want it consistent across all the IRSA resources?

bobdoah:
> One change required and we are good

I have resolved as requested.

modules/kubernetes-addons/keda/variables.tf
```diff
@@ -10,6 +10,12 @@ variable "manage_via_gitops" {
   description = "Determines if the add-on should be managed via GitOps"
 }
 
+variable "iam_role_path" {
+  type        = string
+  default     = "/"
+  description = "IAM role path"
+}
+
```
vara-bonthu, line 18: This variable already exists

vara-bonthu: this needs to be removed

vara-bonthu (changes requested):
One change required and ready to be merged

modules/kubernetes-addons/keda/variables.tf
bobdoah: Removed it now. Didn't see a conflict in the rebase, apologies.

vara-bonthu (approved):
LGTM

PR #356, opened 2022-03-18
Allow usage of depends_on for accelerator module, remove fargate_profiles from base example

Zvikan:
### What does this PR do?
- Create TF resource per  IAM role policy attach for known policies
- Add `depends_on` for `eks-cluster-with-new-vpc` example between accelerator module and vpc
- Remove fargate_profiles from the example to allow GovCloud users to try the example.

### Motivation

Cleaner TF destroy, avoid scenarios where VPC resources are deleted before EKS managed node groups does which leads to exceptions and the need of multiple destroy retries.

### More

- [ ] Yes, I have added a new example under [examples](https://github.com/aws-samples/aws-eks-accelerator-for-terraform/tree/main/examples) to support my PR

### For Moderators

### DIFF
--- a/examples/eks-cluster-with-new-vpc/main.tf
+++ b/examples/eks-cluster-with-new-vpc/main.tf
@@ -121,26 +121,7 @@ module "aws-eks-accelerator-for-terraform" {
       subnet_ids      = module.aws_vpc.private_subnets
     }
   }
-
-  # FARGATE
-  fargate_profiles = {
-    default = {
-      fargate_profile_name = "default"
-      fargate_profile_namespaces = [
-        {
-          namespace = "default"
-          k8s_labels = {
-            Environment = "preprod"
-            Zone        = "dev"
-            env         = "fargate"
-          }
-      }]
-      subnet_ids = module.aws_vpc.private_subnets
-      additional_tags = {
-        ExtraTag = "Fargate"
-      }
-    },
-  }
+  depends_on = [module.aws_vpc]
 }
 
 module "kubernetes-addons" {
--- a/modules/aws-eks-managed-node-groups/locals.tf
+++ b/modules/aws-eks-managed-node-groups/locals.tf
@@ -66,11 +66,7 @@ locals {
     templatefile("${path.module}/templates/userdata-${local.managed_node_group["launch_template_os"]}.tpl", local.userdata_params)
   )
 
-  eks_worker_policies = toset(concat([
-    "${local.policy_arn_prefix}/AmazonEKSWorkerNodePolicy",
-    "${local.policy_arn_prefix}/AmazonEKS_CNI_Policy",
-    "${local.policy_arn_prefix}/AmazonEC2ContainerRegistryReadOnly",
-    "${local.policy_arn_prefix}/AmazonSSMManagedInstanceCore"],
+  eks_worker_policies = toset(concat(
     local.managed_node_group["additional_iam_policies"]
   ))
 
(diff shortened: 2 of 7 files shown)


modules/aws-eks-managed-node-groups/locals.tf
```diff
@@ -66,11 +66,7 @@ locals {
     templatefile("${path.module}/templates/userdata-${local.managed_node_group["launch_template_os"]}.tpl", local.userdata_params)
   )
 
-  eks_worker_policies = toset(concat([
-    "${local.policy_arn_prefix}/AmazonEKSWorkerNodePolicy",
-    "${local.policy_arn_prefix}/AmazonEKS_CNI_Policy",
-    "${local.policy_arn_prefix}/AmazonEC2ContainerRegistryReadOnly",
-    "${local.policy_arn_prefix}/AmazonSSMManagedInstanceCore"],
+  eks_worker_policies = toset(concat(
```
Zvikan, line 69: we probably can remove eks_worker_policies and directly for_each from `local.managed_node_group["additional_iam_policies"]` , but wanted to get your opinion on this

examples/eks-cluster-with-new-vpc/main.tf
```diff
@@ -121,26 +121,7 @@ module "aws-eks-accelerator-for-terraform" {
...
-            env         = "fargate"
-          }
-      }]
-      subnet_ids = module.aws_vpc.private_subnets
-      additional_tags = {
-        ExtraTag = "Fargate"
-      }
-    },
-  }
+  depends_on = [module.aws_vpc]
```
vara-bonthu, line 124: I think we should avoid adding module level dependencies. Implicit dependencies should handle. Moreover, Customers should never try to create VPC and EKS Cluster in one apply. This may be ok for sandbox deployment for dev purposes

modules/aws-eks-managed-node-groups/locals.tf
vara-bonthu: Code looks cleaner this way. We can keep it like that

modules/aws-eks-managed-node-groups/main.tf
```diff
@@ -62,6 +62,9 @@ resource "aws_eks_node_group" "managed_ng" {
   depends_on = [
     aws_iam_role.managed_ng,
     aws_iam_instance_profile.managed_ng,
+    aws_iam_role_policy_attachment.managed_ng_AmazonEKS_CNI_Policy,
+    aws_iam_role_policy_attachment.managed_ng_AmazonEKSWorkerNodePolicy,
+    aws_iam_role_policy_attachment.managed_ng_AmazonEC2ContainerRegistryReadOnly,
```
vara-bonthu, line 67: Needs adding these three dependencies to the launch templates as well

vara-bonthu (changes requested):
I understand this change is to avoid ` Error: Invalid for_each argument` issue with module level dependency using aws.vpc. 

PR looks good to me. Few changes needs to be done. Thanks 👍🏼

examples/eks-cluster-with-new-vpc/main.tf
vara-bonthu: If this is not increasing the TF PLAN time and helping in Destroy then we can keep it

modules/aws-eks-managed-node-groups/main.tf
Zvikan: Done

vara-bonthu (approved):
LGTM 👍🏼

kcoleman731 (approved):

PR #391, opened 2022-04-07
Karpenter fix for AMI and sec group rules

vara-bonthu:
### What does this PR do?

- Adds additional sec group rules to nodes and cluster sec groups as per Karpenter 
- Handling AMI as per the  current region

### More

- [ ] Yes, I have added a new example under [examples](https://github.com/aws-samples/aws-eks-accelerator-for-terraform/tree/main/examples) to support my PR
- [x] Yes, I have updated the [docs](https://github.com/aws-samples/aws-eks-accelerator-for-terraform/tree/main/docs) for this feature

### For Moderators
- [x] E2E Test successfully complete before merge?

### DIFF
--- a/examples/karpenter/main.tf
+++ b/examples/karpenter/main.tf
@@ -147,18 +147,6 @@ module "aws-eks-accelerator-for-terraform" {
   # EKS CONTROL PLANE VARIABLES
   cluster_version = local.cluster_version
 
-  # Extend cluster security group rule to allow Karpenter service to communicate with Cluster API
-  cluster_security_group_additional_rules = {
-    ingress_nodes_karpenter_port = {
-      description                = "From node 8443 for Karpenter"
-      protocol                   = "tcp"
-      from_port                  = 8443
-      to_port                    = 8443
-      type                       = "ingress"
-      source_node_security_group = true
-    }
-  }
-
   # Allow Worker nodes egress to talk to Cluster API for Karpenter service
   node_security_group_additional_rules = {
     egress_nodes_karpenter_port = {
--- a/examples/karpenter/main.tf
+++ b/examples/karpenter/main.tf
@@ -162,11 +162,19 @@ module "aws-eks-accelerator-for-terraform" {
   # Allow Worker nodes egress to talk to Cluster API for Karpenter service
   node_security_group_additional_rules = {
     egress_nodes_karpenter_port = {
-      description                   = "Node groups to cluster API"
+      description                   = "Node groups to cluster API for Karpenter"
       protocol                      = "tcp"
       from_port                     = 8443
       to_port                       = 8443
       type                          = "egress"
+      source_cluster_security_group = true
+    }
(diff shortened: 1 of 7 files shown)


main.tf
```diff
@@ -56,11 +56,14 @@ module "aws_eks" {
...
   # Cluster Security Group
   create_cluster_security_group           = true
-  create_node_security_group              = true
   vpc_id                                  = var.vpc_id
   cluster_additional_security_group_ids   = var.cluster_additional_security_group_ids
   cluster_security_group_additional_rules = var.cluster_security_group_additional_rules
 
+  # Worker Node Security Group
+  create_node_security_group           = true
+  node_security_group_additional_rules = var.node_security_group_additional_rules
```
Zvikan, line 65: Since we have our own node modules, will this work?

vara-bonthu: yes

askulkarni2 (approved):
LGTM!

examples/karpenter/main.tf
```diff
@@ -127,6 +147,37 @@ module "aws-eks-accelerator-for-terraform" {
   # EKS CONTROL PLANE VARIABLES
   cluster_version = local.cluster_version
 
+  # Extend cluster security group rule to allow Karpenter service to communicate with Cluster API
```
bryantbiggs, line 150: feel free to double check me, but I believe we only need the node security group additional rule allowing the control plane to reach the node on 8443 to trigger the Karpenter webhook https://github.com/terraform-aws-modules/terraform-aws-eks/blob/3ff17205a4ead51cca993547ef3de42cc080043b/examples/karpenter/main.tf#L32-L42

vara-bonthu: Thanks I will update this one

vara-bonthu: Just verified and we only need one ingress rule for node security group

vara-bonthu: @bryantbiggs It would be nice to add this rule as a default for upstream module. I think its secured since its only allowing 8443 cluster sec group

bryantbiggs: yes, @Zvikan and I were discussing this as well. I have to think about how best to support or at least show users what ports are required for common services (ALB ingress controller needs 9443/TCP, metrics server needs 10250/tcp and 8443/tcp, etc.)

bryantbiggs (approved):

examples/karpenter/main.tf
vara-bonthu: Yeah, it's not an upstream modules responsibility. It should be done by the Add-on deployment itself. Add-ons should add the necessary rules when its deployed. 

I think we should move this rule from example to the actual add-on itself for better customer experience. We can always output the cluster and worker sec groups from the `eks-blueprints` module and pass these to Kuberentes Add-ons as input. I will not do this change in this PR but we can look at this in our next improvement.

PR #381, opened 2022-04-01
Fix/load balancer controller govcloud support

naris-silpakit:
### What does this PR do?

This PR addresses https://github.com/aws-samples/aws-eks-accelerator-for-terraform/issues/354 by removing hardcoded aws partition values and replacing them with dynamic variables. It also updates the default ECR registry that the load-balancer-controller addon uses so it pulls from the correct region-specific Amazon container image registries outlined [here](https://docs.aws.amazon.com/eks/latest/userguide/add-ons-images.html).

I am keeping this in in draft until I get verification from the issue reporter, @camba1, that these fixes fully resolve the authentication issue on his end. EDIT: Verified working in the issue linked above. Ready for review.

### More

- [x] Yes, I have tested the PR using my local account setup  (Provide any test evidence report under Additional Notes)
- [ ] Yes, I have added a new example under [examples](https://github.com/aws-samples/aws-eks-accelerator-for-terraform/tree/main/examples) to support my PR
- [ ] Yes, I have created another PR for add-ons under [add-ons](https://github.com/aws-samples/ssp-eks-add-ons) repo (if applicable)
- [ ] Yes, I have updated the [docs](https://github.com/aws-samples/aws-eks-accelerator-for-terraform/tree/main/docs) for this feature
- [x] Yes, I ran `pre-commit run -a` with this PR

**Note**: Not all the PRs required examples and docs except a new pattern or add-on added.

### For Moderators
- [ ] E2E Test successfully complete before merge?

### Additional Notes

I ran a full `terraform apply` and `terraform destroy` of `examples/eks-cluster-with-new-vpc` to verify that I did not run into the same auth problem reported in the issue above.

PR #293, opened 2022-02-27
Feature/25 external dns add on

kcoleman731:
### What does this PR do?

This pull request adds support for an [External DNS](https://github.com/kubernetes-sigs/external-dns) add-on.

It also updates `ingress-nginx` add-on to remove boilerplate configuration.

### Motivation

Allow users to leverage External DNS in their EKS clusters.

### More

- [x] Yes, I have tested the PR using my local account setup  (Provide any test evidence report under Additional Notes)
- [x] Yes, I have added a new example under [examples](https://github.com/aws-samples/aws-eks-accelerator-for-terraform/tree/main/examples) to support my PR
- [ ] Yes, I have created another PR for add-ons under [add-ons](https://github.com/aws-samples/ssp-eks-add-ons) repo (if applicable)
- [x] Yes, I have updated the [docs](https://github.com/aws-samples/aws-eks-accelerator-for-terraform/tree/main/docs) for this feature
- [x] Yes, I ran `pre-commit run -a` with this PR

**Note**: Not all the PRs required examples and docs except a new pattern or add-on added.

### For Moderators
- [ ] E2E Test successfully complete before merge?

PR #397, opened 2022-04-10
Fix/ running ArgoCd example on GovCloud

camba1:
### What does this PR do?

PR has two changes to enhance the experience of running the [argocd example](https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/examples/gitops/argocd) in govCloud:

- Remove hardcoded region and move it into a variable that can more easily overridden 
- Update the Nginx module to create the `nginx namespace` instead of the `ingress-nginx namespace`. With the old namespace nothing would get created for Nginx beyond an empty namespace and the following error could be seen on Argo cd: 
`Failed sync attempt to c0d70e7b110803dd967c91fec4cba9bd9622bcab: one or more objects failed to apply, reason: namespaces "nginx" not found (retried 1 times).`
Once the correct namespace is created, everything syncs properly and all appropriate resources are created

### Motivation

Testing the ArgoCd example in govcloud to see if gitops would work as expected.

### Additional Notes

For the ArgoCD example to be fully functional, it requires that [PR 381](https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/381) is passed as well to work properly, without that PR the load balancer will fail to deploy.

### DIFF
--- a/examples/gitops/argocd/main.tf
+++ b/examples/gitops/argocd/main.tf
@@ -175,7 +175,7 @@ module "eks-blueprints-kubernetes-addons" {
   #---------------------------------------------------------------
 
   enable_aws_for_fluentbit            = true
-  enable_aws_load_balancer_controller = false
+  enable_aws_load_balancer_controller = true
   enable_cert_manager                 = true
   enable_cluster_autoscaler           = true
   enable_ingress_nginx                = true
--- a/examples/gitops/argocd/main.tf
+++ b/examples/gitops/argocd/main.tf
@@ -34,7 +34,7 @@ data "aws_eks_cluster_auth" "cluster" {
 }
 
 provider "aws" {
-  region = "us-west-2"
+  region = var.region
   alias  = "default"
 }
 
@@ -175,7 +175,7 @@ module "eks-blueprints-kubernetes-addons" {
   #---------------------------------------------------------------
 
   enable_aws_for_fluentbit            = true
-  enable_aws_load_balancer_controller = true
+  enable_aws_load_balancer_controller = false
   enable_cert_manager                 = true
   enable_cluster_autoscaler           = true
   enable_ingress_nginx                = true
--- a/modules/kubernetes-addons/ingress-nginx/locals.tf
+++ b/modules/kubernetes-addons/ingress-nginx/locals.tf
@@ -1,12 +1,13 @@
 locals {
   name = "ingress-nginx"
+  namespace = "nginx"
 
   default_helm_config = {
     name             = local.name
     chart            = local.name
     repository       = "https://kubernetes.github.io/ingress-nginx"
     version          = "4.0.17"
-    namespace        = local.name
+    namespace        = local.namespace
     create_namespace = false
     values           = local.default_helm_values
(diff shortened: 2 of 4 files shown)


examples/gitops/argocd/main.tf
```diff
@@ -34,7 +34,7 @@ data "aws_eks_cluster_auth" "cluster" {
 }
 
 provider "aws" {
-  region = "us-west-2"
+  region = var.region
```
vara-bonthu, line 37: Like that. Should be done for all examples.

vara-bonthu (approved):
Thanks @camba1 👍🏼  LGTM

modules/kubernetes-addons/ingress-nginx/locals.tf
```diff
@@ -1,12 +1,13 @@
 locals {
   name = "ingress-nginx"
+  namespace = "nginx"
```
askulkarni2, line 3: Why are we changing this? Should be `ingress-nginx` as per https://kubernetes.github.io/ingress-nginx/deploy/#quick-start.

askulkarni2 (commented):
@camba1 thanks for the PR. Question on the changing namespace.

modules/kubernetes-addons/ingress-nginx/locals.tf
camba1: Good question @askulkarni2. Per the [add-on documentation](https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/docs/add-ons/nginx.md):

"Note that the ingress controller is deployed in the **nginx** namespace." 

So, it appears that there must be a reason for this to be done this way.

This is consistent with the error that was seen when deploy the add-on  before the change:

`Failed sync attempt to c0d70e7b110803dd967c91fec4cba9bd9622bcab: one or more objects failed to apply, reason: namespaces "nginx" not found (retried 1 times).`

 In fact this bug/change was introduced recently when PR #293 was merged 4 days ago. Before then the value for the namespace was indeed 'nginx' and not 'ingress-nginx'

askulkarni2: To me recollection, the error you are referring to was a mis-configuration of the addons repo. Is there an error you are seeing with `ingress-nginx`. Because again as per official documentation of the addon the namepsace is `ingress-nginx`. We will need to fix our documentation to reflect that.

kcoleman731: @camba1 I believe the place we need to update namespace is here https://github.com/aws-samples/eks-blueprints-add-ons/blob/main/chart/templates/ingress-nginx.yaml#L17

camba1: Updated the repo above with the new namespace [PR](https://github.com/aws-samples/eks-blueprints-add-ons/pull/30) . Removed changes to the nginx locals and updated the doc to reflect the new namespace in this repo.
Please let me know if here is anything else that needs to be done

askulkarni2 (approved):
LGTM!

PR #398, opened 2022-04-11
Ensure serviceaccount created in correct namespace

bobdoah:
### What does this PR do?

The serviceaccount for a module is created via the IRSA module. If the
namespace of a module differs from the default, it needs to also be
passed to the IRSA module. Otherwise any deployments tied to the
serviceaccount will fail to deploy pods. 

This PR changes the modules so that they can be deployed in a namespace other than "kube-system".

### Motivation

 I deploy the aws-load-balancer to a different namespace, in my configuration. It is a security requirement. I am unable to use recent versions of this module to achieve that. This should fix that issue.

### More

### For Moderators

### DIFF
--- a/modules/kubernetes-addons/aws-load-balancer-controller/locals.tf
+++ b/modules/kubernetes-addons/aws-load-balancer-controller/locals.tf
@@ -41,7 +41,7 @@ locals {
   }
 
   irsa_config = {
-    kubernetes_namespace              = "kube-system"
+    kubernetes_namespace              = local.helm_config["namespace"]
     kubernetes_service_account        = local.service_account_name
     create_kubernetes_namespace       = false
     create_kubernetes_service_account = true
--- a/modules/kubernetes-addons/aws-load-balancer-controller/locals.tf
+++ b/modules/kubernetes-addons/aws-load-balancer-controller/locals.tf
@@ -40,16 +40,11 @@ locals {
     serviceAccountName = local.service_account_name
   }
 
-  default_irsa_config = {
+  irsa_config = {
     kubernetes_namespace              = local.helm_config["namespace"]
     kubernetes_service_account        = local.service_account_name
-    create_kubernetes_namespace       = var.kubernetes_namespace != "kube-system" ? true : false
+    create_kubernetes_namespace       = try(local.helm_config["create_namespace"], true)
     create_kubernetes_service_account = true
     irsa_iam_policies                 = [aws_iam_policy.aws_load_balancer_controller.arn]
   }
-
-  irsa_config = merge(
-    local.default_irsa_config,
-    var.irsa_config
-  )
 }
--- a/modules/kubernetes-addons/aws-load-balancer-controller/locals.tf
+++ b/modules/kubernetes-addons/aws-load-balancer-controller/locals.tf
@@ -40,11 +40,16 @@ locals {
     serviceAccountName = local.service_account_name
   }
 
-  irsa_config = {
+  default_irsa_config = {
     kubernetes_namespace              = local.helm_config["namespace"]
     kubernetes_service_account        = local.service_account_name
-    create_kubernetes_namespace       = false
+    create_kubernetes_namespace       = var.kubernetes_namespace != "kube-system" ? true : false
     create_kubernetes_service_account = true
     irsa_iam_policies                 = [aws_iam_policy.aws_load_balancer_controller.arn]
   }
+
+  irsa_config = merge(
+    local.default_irsa_config,
+    var.irsa_config
+  )
 }
--- a/modules/kubernetes-addons/aws-load-balancer-controller/variables.tf
+++ b/modules/kubernetes-addons/aws-load-balancer-controller/variables.tf
@@ -27,9 +27,3 @@ variable "addon_context" {
   })
   description = "Input configuration for the addon."
 }
-
-variable "irsa_config" {
-  type        = any
-  description = "IRSA config for the aws_load_balancer_controller. "
-  default     = {}
-}
--- a/modules/kubernetes-addons/aws-load-balancer-controller/variables.tf
+++ b/modules/kubernetes-addons/aws-load-balancer-controller/variables.tf
@@ -27,3 +27,9 @@ variable "addon_context" {
   })
   description = "Input configuration for the addon."
 }
+
+variable "irsa_config" {
+  type        = any
+  description = "IRSA config for the aws_load_balancer_controller. "
+  default     = {}
+}
--- a/modules/kubernetes-addons/cluster-autoscaler/locals.tf
+++ b/modules/kubernetes-addons/cluster-autoscaler/locals.tf
@@ -36,7 +36,7 @@ locals {
 
   irsa_config = {
     create_kubernetes_namespace       = false
-    kubernetes_namespace              = local.helm_config["namespace"]
+    kubernetes_namespace              = "kube-system"
     create_kubernetes_service_account = true
     kubernetes_service_account        = local.service_account_name
     irsa_iam_policies                 = [aws_iam_policy.cluster_autoscaler.arn]
--- a/modules/kubernetes-addons/cluster-autoscaler/locals.tf
+++ b/modules/kubernetes-addons/cluster-autoscaler/locals.tf
@@ -36,7 +36,7 @@ locals {
 
   irsa_config = {
     create_kubernetes_namespace       = false
-    kubernetes_namespace              = "kube-system"
+    kubernetes_namespace              = local.helm_config["namespace"]
     create_kubernetes_service_account = true
     kubernetes_service_account        = local.service_account_name
     irsa_iam_policies                 = [aws_iam_policy.cluster_autoscaler.arn]
--- a/docs/add-ons/aws-cloudwatch-metrics.md
+++ b/docs/add-ons/aws-cloudwatch-metrics.md
@@ -1,0 +1,40 @@
+# AWS CloudWatch Metrics
+
+Use CloudWatch Container Insights to collect, aggregate, and summarize metrics and logs from your containerized applications and microservices. CloudWatch automatically collects metrics for many resources, such as CPU, memory, disk, and network. Container Insights also provides diagnostic information, such as container restart failures, to help you isolate issues and resolve them quickly. You can also set CloudWatch alarms on metrics that Container Insights collects.
+
+Container Insights collects data as performance log events using embedded metric format. These performance log events are entries that use a structured JSON schema that enables high-cardinality data to be ingested and stored at scale. From this data, CloudWatch creates aggregated metrics at the cluster, node, pod, task, and service level as CloudWatch metrics. The metrics that Container Insights collects are available in CloudWatch automatic dashboards, and also viewable in the Metrics section of the CloudWatch console.
+
+## Usage
+
+[aws-cloudwatch-metrics](https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/modules/kubernetes-addons/aws-cloudwatch-metrics) can be deployed by enabling the add-on via the following.
+
+```hcl
+enable_aws_cloudwatch_metrics = true
+```
+
+You can optionally customize the Helm chart that deploys `aws_cloudwatch_metrics` via the following configuration.
+
+```hcl
+  enable_aws_cloudwatch_metrics        = true
+  aws_cloudwatch_metrics_irsa_policies = ["IAM Policies"]
+  aws_cloudwatch_metrics_helm_config   = {
+    name       = "aws-cloudwatch-metrics"
+    chart      = "aws-cloudwatch-metrics"
+    repository = "https://aws.github.io/eks-charts"
+    version    = "0.0.7"
+    namespace  = "amazon-cloudwatch"
+    values     = [templatefile("${path.module}/values.yaml", {
+      eks_cluster_id = var.addon_context.eks_cluster_id
+    })]
+  }
+```
+
+### GitOps Configuration
+
+The following properties are made available for use when managing the add-on via GitOps.
+
+```hcl
+awsCloudWatchMetrics = {
+  enable       = true
+}
+```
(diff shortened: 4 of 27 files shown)


askulkarni2 (approved):
@bobdoah thanks for the PR!  LGTM!

modules/kubernetes-addons/aws-load-balancer-controller/locals.tf
```diff
@@ -40,7 +40,7 @@ locals {
   }
 
   irsa_config = {
-    kubernetes_namespace              = "kube-system"
+    kubernetes_namespace              = local.helm_config["namespace"]
     kubernetes_service_account        = local.service_account_name
     create_kubernetes_namespace       = false
```
vara-bonthu, line 45: This needs to be changed to `create_kubernetes_namespace = true`  when we use namespace other than `kube-system` to ensure namespace is created by the IRSA module

vara-bonthu: @askulkarni2 May be we need to handle this in IRSA module to look for `kube-system`  when `create_kubernetes_namespace=true`.

See below 

```hcl
resource "kubernetes_namespace_v1" "irsa" {
  count = var.create_kubernetes_namespace && var.kubernetes_namespace != "kube-system" ? 1 : 0
  metadata {
    name = var.kubernetes_namespace

    labels = {
      "app.kubernetes.io/managed-by" = "terraform-eks-blueprints"
    }
  }
}
```

vara-bonthu (changes requested):
Thanks for the PR 👍🏼  Few changes required

modules/kubernetes-addons/aws-load-balancer-controller/locals.tf
bobdoah: > This needs to be changed to `create_kubernetes_namespace = true` when we use namespace other than `kube-system` to ensure namespace is created by the IRSA module

I'm not sure I follow the logic. In my use-case, the namespace already exists. So this wouldn't work for me. That isn't done in other instances such as for the [kubernetes_dashboard module](https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/f4a1a00bd92f5fe2e7cd44d6883cd90d291d1f9b/modules/kubernetes-addons/kubernetes-dashboard/locals.tf#L38). 

I would suggest the `create_kubernetes_namespace` option should be configurable for each of the given addon modules. Perhaps as part of the [`addon_context`](https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/f4a1a00bd92f5fe2e7cd44d6883cd90d291d1f9b/modules/kubernetes-addons/aws-load-balancer-controller/variables.tf#L13). 

I'd be happy to look at that, but this is not the intent of this PR. I'm trying to draw these modules level with others such as the `kubernetes_dashboard` that would work with an existing namespace that is _not_ `kube-system`. This behaviour did previously work.

vara-bonthu: Thanks @bobdoah I appreciate your logic. Let me go through in detail

Namespace creation is handled only by EKS Blueprints module at this moment apart from existing system namespaces like `kube-system`. Users can pass any name as namespace through the below configuration. 

```hcl
  enable_aws_load_balancer_controller = true
  # Optional  
  aws_load_balancer_controller_helm_config = {
    name                       = "aws-load-balancer-controller"
    chart                      = "aws-load-balancer-controller"
    repository                 = "https://aws.github.io/eks-charts"
    version                    = "1.3.1"
    namespace                  = "my_alb_ns"
    values = [templatefile("${path.module}/values.yaml", {})]
  }
```
User defined `my_alb_ns` namespace overrides the locally defined namespace. When we specify `kubernetes_namespace = local.helm_config["namespace"]` and `create_kubernetes_namespace = false`  then the above example will introduce a defect as `my_alb_ns` wont be created due to `false` condition. 

Yes, I just noticed that  `kubernetes-dashboard` config is incorrect as well. I think the solution for this would be as follows

- We will discuss about your suggestion allowing users to define `create_kubernetes_namespace` through helm config

bobdoah: @vara-bonthu I understand what you're saying, but that doesn't fit my use case. The namespace has already been created externally, it's not `kube-system`. I have no control over this, sadly. If I set it to `true` I still won't be able to use this module, as-is. 

I have pushed a further commit. I think this is the simplest way I can satisfy both yours and my use cases. The namespace will be created, if it is not the default `kube-system`. The `create_kubernetes_namespace` option can be overridden by supplying a custom `irsa_config` to the module. The contents will be merged, in the same way as is done for the `helm_config`.

kcoleman731 (approved):
LGTM. @vara-bonthu anything else needed here?

modules/kubernetes-addons/aws-load-balancer-controller/locals.tf
```diff
@@ -39,11 +39,16 @@ locals {
...
-    create_kubernetes_namespace       = false
+    create_kubernetes_namespace       = var.kubernetes_namespace != "kube-system" ? true : false
     create_kubernetes_service_account = true
     irsa_iam_policies                 = [aws_iam_policy.aws_load_balancer_controller.arn]
   }
+
+  irsa_config = merge(
+    local.default_irsa_config,
+    var.irsa_config
+  )
```
vara-bonthu, line 53: remove this block

modules/kubernetes-addons/aws-load-balancer-controller/locals.tf
```diff
@@ -39,11 +39,16 @@ locals {
     serviceAccountName = local.service_account_name
   }
 
-  irsa_config = {
-    kubernetes_namespace              = "kube-system"
+  default_irsa_config = {
+    kubernetes_namespace              = local.helm_config["namespace"]
     kubernetes_service_account        = local.service_account_name
-    create_kubernetes_namespace       = false
+    create_kubernetes_namespace       = var.kubernetes_namespace != "kube-system" ? true : false
```
vara-bonthu, line 45: replace this with 
```hcl
create_kubernetes_namespace  = try(local.helm_config["create_namespace"], true)
```

modules/kubernetes-addons/aws-load-balancer-controller/locals.tf
```diff
@@ -39,11 +39,16 @@ locals {
     serviceAccountName = local.service_account_name
   }
 
-  irsa_config = {
-    kubernetes_namespace              = "kube-system"
+  default_irsa_config = {
```
vara-bonthu, line 42: change this back to `irsa_config`

modules/kubernetes-addons/aws-load-balancer-controller/variables.tf
```diff
@@ -26,3 +26,9 @@ variable "addon_context" {
   })
   description = "Input configuration for the addon"
 }
+
+variable "irsa_config" {
+  type        = any
+  description = "IRSA config for the aws_load_balancer_controller. "
+  default     = {}
+}
```
vara-bonthu, line 34: remove this variable

modules/kubernetes-addons/aws-load-balancer-controller/locals.tf
vara-bonthu: replace this line in  **IRSA module**  https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/91dc3651d33049029f7b4188b515ab16a1b11a24/modules/irsa/main.tf#L2

with the below

```hcl
count = var.create_kubernetes_namespace && var.kubernetes_namespace != "kube-system" ? 1 : 0
```

modules/kubernetes-addons/cluster-autoscaler/locals.tf
```diff
@@ -36,7 +36,7 @@ locals {
 
   irsa_config = {
     create_kubernetes_namespace       = false
-    kubernetes_namespace              = "kube-system"
```
vara-bonthu, line 39: Here we are forcing the Cluster CA to run on `kube-ssytem`.

Running CA in other namespace on a user node is risky. If CA erases the node it is sitting on and some race condition occurs you may end up with cluster without both CA and capacity to run anything. 

Hence its really important that...

CA stays in `kube-system` OR it has restrictive PodDisruptionBudget
CA is run at high priority or is marked as a critical pod

vara-bonthu (changes requested):
Hi @bobdoah I thought about a better way of handling this scenario. Please see my new  comments to the PR. 

Please follow these steps if you want to bring your own namespace.

**Note:** 
- EKS Blueprints manages the Namespace by default
- If user wants to manage your namespace then you can pass the following config.

```hcl
  #---------------------------------------
  # AWS LOAD BALANCER INGRESS CONTROLLER HELM ADDON
  #---------------------------------------
  enable_aws_load_balancer_controller = true
  # Optional
  aws_load_balancer_controller_helm_config = {
    name       = "aws-load-balancer-controller"
    namespace  = "my-own-namespace"
    create_namespace = false
  }
```


Give this change a try and let me know if it works for you. I will try to update the other Add-ons with the similar approach.

modules/kubernetes-addons/aws-load-balancer-controller/locals.tf
vara-bonthu, line 45:
```suggestion
    create_kubernetes_namespace  = try(local.helm_config["create_namespace"], true)
```

bobdoah:
@vara-bonthu updated as requested.

vara-bonthu (approved):

ZeroDeth:
Facing the same issue with multiple add-ons:
- tetrate_istio
- cert-manager
- traefik
- aws_load_balancer_controller

bobdoah:
> Facing the same issue with multiple add-ons: `pending-installs`
> 
> * tetrate_istio
> * cert-manager
> * traefik
> * aws_load_balancer_controller

Applied the same try/true default to cert-manager.

The `tetrate_istio` module is a completely separate repository. It does appear to support a `create_namespace` argument on the Helm config https://github.com/tetratelabs/terraform-eksblueprints-tetrate-istio-addon/blob/5027357081f0e62dc8fb0c135197c908b6463e4b/locals.tf#L13. @ZeroDeth, I think you should be able to make that work by supplying:
```
module "kubernetes_addons" {
 source = "github.com/aws-ia/terraform-aws-eks-blueprints//modules/kubernetes-addons"
  
  ...
  tetrate_istio_base_helm_config = {
    create_namespace = false
  }
}
```   
The traefik module is more complicated, as the namespace resource is created directly in the module. It could be put behind a `count` like so:

```
 resource "kubernetes_namespace_v1" "this" {
+  count = try(local.helm_config["create_namespace"], true) ? 1 : 0
   metadata {
     name = local.helm_config["namespace"]
```

If that would suffice. It would, however potentially remove the resources of anyone upgrading. A `moved` block would help with this, but that's only supported in Terraform >1.1.

bobdoah:
@vara-bonthu @kcoleman731 @askulkarni2 any chance of getting this merged, now that @vara-bonthu is happy.

Thanks

vara-bonthu:
@bobdoah Happy to merge now. Just checking to see if this PR is working for you?

bobdoah:
@vara-bonthu Yep, this is working for me.

ZeroDeth:
`cert-manager` & `ingress-nginx` failed now. probably I am doing it wrong.

```sh
Warning: Helm release "cert-manager" was created but has a failed status. Use the `helm` command to investigate the error, correct it, then run Terraform again.

with module.eks_addons_0.module.cert_manager[0].module.helm_addon.helm_release.addon[0], on .terraform/modules/eks_addons_0/modules/kubernetes-addons/helm-addon/main.tf line 1, in resource "helm_release" "addon":1: resource "helm_release" "addon" {

Warning: Helm release "ingress-nginx" was created but has a failed status. Use the `helm` command to investigate the error, correct it, then run Terraform again.

with module.eks_addons_0.module.ingress_nginx[0].module.helm_addon.helm_release.addon[0], on .terraform/modules/eks_addons_0/modules/kubernetes-addons/helm-addon/main.tf line 1, in resource "helm_release" "addon":1: resource "helm_release" "addon" {

Error: failed post-install: timed out waiting for the condition with module.eks_addons_0.module.cert_manager[0].module.helm_addon.helm_release.addon[0], on .terraform/modules/eks_addons_0/modules/kubernetes-addons/helm-addon/main.tf line 1, in resource "helm_release" "addon":1: resource "helm_release" "addon" {

Error: timed out waiting for the condition with module.eks_addons_0.module.ingress_nginx[0].module.helm_addon.helm_release.addon[0], on .terraform/modules/eks_addons_0/modules/kubernetes-addons/helm-addon/main.tf line 1, in resource "helm_release" "addon":1: resource "helm_release" "addon" {
```

example 1
```sh
  enable_cert_manager = true
  cert_manager_helm_config = {
    name             = "cert-manager"
    chart            = "cert-manager"
    repository       = "https://charts.jetstack.io"
    version          = "v1.7.2"
    namespace        = "cert-manager"
    create_namespace = true
  }

  enable_ingress_nginx = true
  ingress_nginx_helm_config = {
    name       = "ingress-nginx"
    chart      = "ingress-nginx"
    repository = "https://kubernetes.github.io/ingress-nginx"
    version    = "4.0.17"
    values = [templatefile("${path.module}/helm_values/nginx-values.yaml", {
      hostname     = var.hosted_name
      ssl_cert_arn = module.shr_acm.acm_certificate_arn
    })]
  }
```

example 2
```sh
  enable_cert_manager = true

  enable_ingress_nginx = true
  ingress_nginx_helm_config = {
    values = [templatefile("${path.module}/helm_values/nginx-values.yaml", {
      hostname     = var.hosted_name
      ssl_cert_arn = module.shr_acm.acm_certificate_arn
    })]
  }
```

ZeroDeth:
> > Facing the same issue with multiple add-ons: `pending-installs`
> > 
> > * tetrate_istio
> > * cert-manager
> > * traefik
> > * aws_load_balancer_controller
> 
> Applied the same try/true default to cert-manager.
> 
> The `tetrate_istio` module is a completely separate repository. It does appear to support a `create_namespace` argument on the Helm config https://github.com/tetratelabs/terraform-eksblueprints-tetrate-istio-addon/blob/5027357081f0e62dc8fb0c135197c908b6463e4b/locals.tf#L13. @ZeroDeth, I think you should be able to make that work by supplying:
> 
> ```
> module "kubernetes_addons" {
>  source = "github.com/aws-ia/terraform-aws-eks-blueprints//modules/kubernetes-addons"
>   
>   ...
>   tetrate_istio_base_helm_config = {
>     create_namespace = false
>   }
> }
> ```
> 
> The traefik module is more complicated, as the namespace resource is created directly in the module. It could be put behind a `count` like so:
> 
> ```
>  resource "kubernetes_namespace_v1" "this" {
> +  count = try(local.helm_config["create_namespace"], true) ? 1 : 0
>    metadata {
>      name = local.helm_config["namespace"]
> ```
> 
> If that would suffice. It would, however potentially remove the resources of anyone upgrading. A `moved` block would help with this, but that's only supported in Terraform >1.1.

Same errors with tetrate_istio

PR #401, opened 2022-04-12
chore: nginx example to leverage aws load balancer controller

Zvikan:
### What does this PR do?

- remove default values from nginx addon module
- enable AWS load balancer controller (LBC) in nginx example
- LBC to be deployed before nginx, and to be destroyed after nginx to avoid issues that resources are left behind and nginx can't delete namespace properly - see https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1629
- Adjust example's README 

### Motivation

- nginx example failing right now using default annotations of LBC while LBC not being deployed
- Show correct order of deployments instead of single TF apply/destroy to avoid issues where VPC resources deleted before addons/cluster, and for nginx to be deleted before LBC for this example.

### More

### For Moderators

### DIFF
--- a/modules/kubernetes-addons/ingress-nginx/values.yaml
+++ b/modules/kubernetes-addons/ingress-nginx/values.yaml
@@ -2,10 +2,8 @@ controller:
   service:
     externalTrafficPolicy: "Local"
     annotations:
-      # AWS Load Balancer Controller Annotations
-      service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp # or 'ssl'
-      service.beta.kubernetes.io/aws-load-balancer-attributes: load_balancing.cross_zone.enabled=true
+      service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp
       service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: '60'
-      service.beta.kubernetes.io/aws-load-balancer-type: 'external'
-      service.beta.kubernetes.io/aws-load-balancer-scheme: 'internet-facing' # or 'internal'
-      service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: 'ip'
+      service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled: 'true'
+      service.beta.kubernetes.io/aws-load-balancer-internal: "false"
+      service.beta.kubernetes.io/aws-load-balancer-type: "nlb"
--- a/modules/kubernetes-addons/ingress-nginx/values.yaml
+++ b/modules/kubernetes-addons/ingress-nginx/values.yaml
@@ -1,9 +1,0 @@
-controller:
-  service:
-    externalTrafficPolicy: "Local"
-    annotations:
-      service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp
-      service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: '60'
-      service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled: 'true'
-      service.beta.kubernetes.io/aws-load-balancer-internal: "false"
-      service.beta.kubernetes.io/aws-load-balancer-type: "nlb"
--- a/.github/workflows/e2e-parallel-destroy.yml
+++ b/.github/workflows/e2e-parallel-destroy.yml
@@ -49,6 +49,14 @@ jobs:
             tenant_name: "agones"
           - example_path: examples/ingress-controllers/nginx
             tenant_name: "nginx"
+            deployment_order:
+              [
+                "module.e2e-test.module.aws_vpc",
+                "module.e2e-test.module.eks-blueprints",
+                "module.e2e-test.module.eks-blueprints-kubernetes-addons",
+                "module.e2e-test.module.aws_load_balancer_controller",
+                "module.e2e-test.module.ingress_nginx",
+              ]
           - example_path: examples/karpenter
             tenant_name: "karpenter"
           - example_path: examples/node-groups/managed-node-groups
--- a/.github/workflows/e2e-parallel-full.yml
+++ b/.github/workflows/e2e-parallel-full.yml
@@ -53,6 +53,14 @@ jobs:
             tenant_name: "agones"
           - example_path: examples/ingress-controllers/nginx
             tenant_name: "nginx"
+            deployment_order:
+              [
+                "module.e2e-test.module.aws_vpc",
+                "module.e2e-test.module.eks-blueprints",
+                "module.e2e-test.module.eks-blueprints-kubernetes-addons",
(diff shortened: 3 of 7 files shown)


kcoleman731 (approved):

kcoleman731 (approved):
LGTM

bryantbiggs (approved):

vara-bonthu (approved):
LGTM 👍🏼  Assume this has been tested

modules/kubernetes-addons/ingress-nginx/values.yaml
```diff
@@ -2,10 +2,8 @@ controller:
...
-      service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp # or 'ssl'
-      service.beta.kubernetes.io/aws-load-balancer-attributes: load_balancing.cross_zone.enabled=true
+      service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp
       service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: '60'
-      service.beta.kubernetes.io/aws-load-balancer-type: 'external'
-      service.beta.kubernetes.io/aws-load-balancer-scheme: 'internet-facing' # or 'internal'
-      service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: 'ip'
+      service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled: 'true'
+      service.beta.kubernetes.io/aws-load-balancer-internal: "false"
+      service.beta.kubernetes.io/aws-load-balancer-type: "nlb"
```
vara-bonthu, line 9: Should this be `service.beta.kubernetes.io/aws-load-balancer-type: nlb-ip` ?

modules/kubernetes-addons/ingress-nginx/values.yaml
```diff
@@ -2,10 +2,8 @@ controller:
...
-      # AWS Load Balancer Controller Annotations
-      service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp # or 'ssl'
-      service.beta.kubernetes.io/aws-load-balancer-attributes: load_balancing.cross_zone.enabled=true
+      service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp
       service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: '60'
-      service.beta.kubernetes.io/aws-load-balancer-type: 'external'
-      service.beta.kubernetes.io/aws-load-balancer-scheme: 'internet-facing' # or 'internal'
-      service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: 'ip'
+      service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled: 'true'
+      service.beta.kubernetes.io/aws-load-balancer-internal: "false"
```
vara-bonthu, line 8: This annotation is now deprecated. You might need to use this instead `service.beta.kubernetes.io/aws-load-balancer-scheme: internal` 

Valid values are `internal`, `internet-facing`

modules/kubernetes-addons/ingress-nginx/values.yaml
```diff
@@ -2,10 +2,8 @@ controller:
   service:
     externalTrafficPolicy: "Local"
     annotations:
-      # AWS Load Balancer Controller Annotations
-      service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp # or 'ssl'
-      service.beta.kubernetes.io/aws-load-balancer-attributes: load_balancing.cross_zone.enabled=true
```
vara-bonthu, line 7: This seems to be correct

modules/kubernetes-addons/ingress-nginx/values.yaml
```diff
@@ -2,10 +2,8 @@ controller:
...
+      service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp
       service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: '60'
-      service.beta.kubernetes.io/aws-load-balancer-type: 'external'
-      service.beta.kubernetes.io/aws-load-balancer-scheme: 'internet-facing' # or 'internal'
-      service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: 'ip'
+      service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled: 'true'
```
vara-bonthu, line 7: This is deprecated annotation and the new one is `service.beta.kubernetes.io/aws-load-balancer-attributes: load_balancing.cross_zone.enabled=true`

modules/kubernetes-addons/ingress-nginx/values.yaml
```diff
@@ -2,10 +2,8 @@ controller:
   service:
     externalTrafficPolicy: "Local"
     annotations:
-      # AWS Load Balancer Controller Annotations
-      service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp # or 'ssl'
-      service.beta.kubernetes.io/aws-load-balancer-attributes: load_balancing.cross_zone.enabled=true
+      service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp
       service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: '60'
```
vara-bonthu, line 6: This annotation is not available in the latest AWS LB Controller code

kcoleman731 (commented):
LGTM

Tstrul-work:
Hello,

I wonder why we need to install both alb controller with nginx and why before this pr we didn't?
as far as I know, Nginx chart can create NLB if we create a service account and assign IAM role, so what is the benefit of installing both of them?

Zvikan:
Hey @Tstrul-work , 
Before this PR, the nginx example didn't work as there was another [change](https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/d110446deca55bfa3ae662a1216dacffb6ac0a22/modules/kubernetes-addons/ingress-nginx/values.yaml) to the nginx controller which used annotations of AWS Load Balancer Controller (LBC) , this was mistake that we wanted to address.

We reviewed the nginx ingress controller [documentation](https://kubernetes.github.io/ingress-nginx/deploy/#aws) to understand what should be our next steps, and we followed the documentation recommendation of using LBC.

Some customers may still use nginx as defined "legacy" in their docs with their annotations, other may choose to use LBC only (as they don't need NGINX at all), we wanted to see if a hybrid approach can work for both sides without confusing too much (as some annotations are the same, some are different).

Your feedback is important, if you think we should separate this examples (e.g. one only for nginx "legacy" , one for LBC only etc.)  or any other way to make this example better, please let us know! thanks.

Tstrul-work:
Hey,
Sorry for the late response and thank you for your answer!

Actually your explenation was perfect, i didnt know about the changes to ingress controller (from the info box)

One improvement  to consider:
We are using two ingress controllers with two ingress class (external and internal) i dont htink this type of configuration is supported in this module

Again,
Thanks a lot for your help!

PR #403, opened 2022-04-12
Apache License update for AWS IA Org Modules 

vara-bonthu:
### What does this PR do?

- Updated License as per AWS-IA module requirement
- Added NOTICE file as per AWS-IA module requirement

### More


tonynv (approved):

LICENSE
```diff
@@ -1,14 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
bryantbiggs, line 189: Should we update or remove? I see some other projects have this as well like Karpenter - @spotaws do you have any input?

vara-bonthu: I think the NOTICE file that we added below covers the Copyright Owners section.

kcoleman731 (approved):

LICENSE
spotaws: I think it is fine to leave the APPENDIX intact (and generic). It is also fine to omit it. I have a slight preference towards leaving it as is, because it makes it easier for license checking tools to identify it as unmodified Apache 2.0.

PR #424, opened 2022-04-19
docs: Update and standardize examples to follow conventions defined in pre-commit checks

### LINKED ISSUE #469: [QUESTION] Licence MIT-0 vs. Apache 2.0
#### Which licence is now used for this repository ? 

The Readme.md states that MIT-0 is used as licence but the referenced file contains the Apache 2.0 Licence

#### Provide link to the example related to the question
https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/README.md#license
vs.
https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/LICENSE

#### More

- [x] Yes, I have checked the repo for existing issues before raising this question

bryantbiggs:
### What does this PR do?
- Standardize AWS provider region usage within examples; remove use of variable for region in provider since this should be static
- Remove references to prior license material and MIT licensing
- Standardize provider and required versions within `versions.tf` per conventions; this also aligns versions that are used to be consistent
- Update pre-commit checks
	- Update versions to use latest
	- Remove `.tflint.hcl` file that wasn't enforcing lint rules, add rules to pre-commit to ensure conventions are enforced
	- Update `terraform_docs` to ensure lockfile doesn't affect version shown in READMEs
- Update and align version used on VPC module within examples
- Move outputs to `outputs.tf` file per convention; add empty file where one is not present in the directory
- Correct tflint errors
	- Remove  unused locals and data sources
	- Ensure variables and outputs have descriptions
	- Ensure names use underscores instead of hyphens
	- The managed node group policy attachment resource names are left as is with a `TODO` for now since that is a breaking change  
	- Ensure provider versions are set in required providers
	- Quite a few whitepsace changes
- Correct `terraform_docs` replacement pattern so that pre-commit will update documentation on detected changes; this was not updating previously and the documentation was out of sync with the source
- Update `kubernetes`, `helm`, and `kubectl` providers to use `exec()` method per the Terraform provider docs. This follows same pattern used by the EKS service within the `kubeconfig` file to ensure credentials are refreshed. This also removes the use of two data sources `aws_eks_cluster_auth` and `aws_eks_cluster` per example

### Motivation
- Standardize and align with current conventions set within static analysis tools
- All pre-commit checks pass successfully
- Closes #469 

### More

### For Moderators


deploy/e2e/eks/README.md
```diff
@@ -10,62 +10,6 @@
...
     terraform apply -var-file base.tfvars -auto-approve
 
-
 ## How to Destroy the cluster
 
-    terraform destroy -var-file base.tfvars -auto-approve  
-
-
-
-<!--- BEGIN_TF_DOCS --->
```
Zvikan, line 20: I believe the terraform docs check will re-add this, unless something changed on the check that I might have missed.

bryantbiggs: ah, this is the thing you are referring to I assume https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/34921232356fb69b5a41094246d488c37c596d97/.github/workflows/terraform-docs.yaml#L26-L33

looks like I'll have to update the `template` param so that remote and local both work for formatting docs - thanks!

bryantbiggs: addressed in https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/424/commits/19a5f13a8a21e9640bb8035be6b3d116c51fc5ee

this will most likely fail first run if merged as is (action runs on `main`, change is in PR), but should work there after

examples/analytics/emr-on-eks/main.tf
```diff
@@ -1,62 +1,40 @@
...
-  }
+provider "aws" {
+  region = local.region
 }
 
-provider "aws" {}
-
 provider "kubernetes" {
-  experiments {
-    manifest_resource = true
```
Zvikan, line 28: we don't need this anymore?

bryantbiggs: I don't believe so, no - this has been available since 2.4 and it looked like the lowest version we had set was 2.7

https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/guides/alpha-manifest-migration-guide

Also, I don't think we even have any `kubernetes_manifest` resources in use in the repo

deploy/pr/main.tf
```diff
@@ -45,5 +17,4 @@ module "eks-cluster-with-import-vpc" {
 
   vpc_id             = data.terraform_remote_state.vpc_s3_backend.outputs.vpc_id
   private_subnet_ids = data.terraform_remote_state.vpc_s3_backend.outputs.private_subnets
-  public_subnet_ids  = data.terraform_remote_state.vpc_s3_backend.outputs.public_subnets
```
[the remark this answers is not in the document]
bryantbiggs: Terraform validate gave this error when it was present: 

```
│ Error: Unsupported argument
│ 
│   on main.tf line 20, in module "eks_cluster_with_import_vpc":
│   20:   public_subnet_ids  = data.terraform_remote_state.vpc_s3_backend.outputs.public_subnets
│ 
│ An argument named "public_subnet_ids" is not expected here.
```

modules/kubernetes-addons/main.tf
```diff
@@ -268,7 +268,6 @@ module "yunikorn" {
   count             = var.enable_yunikorn ? 1 : 0
   source            = "./yunikorn"
   helm_config       = var.yunikorn_helm_config
-  irsa_policies     = var.yunikorn_irsa_policies
```
Zvikan, line 271: why we removing this? `irsa_policies` not being used at all in the module?

bryantbiggs: no sir - not in use

Zvikan (approved):

deploy/pr/main.tf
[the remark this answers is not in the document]
Zvikan: cc @vara-bonthu , do we still need PR, are we still using pr-test?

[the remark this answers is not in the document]
vara-bonthu: Actually, you can remove the `deploy/pr`  folder and `pr-test.yaml` github workflow. We can replace this with `examples-tfplan-tests.yaml` github workflow.

deploy/pr/versions.tf
```diff
@@ -0,0 +1,24 @@
...
+      version = ">= 2.4.1"
+    }
+  }
+
+  backend "s3" {
+    bucket = "terraform-ssp-github-actions-state"
+    key    = "pr/eks/terraform-main.tfstate"
+    region = "us-west-2"
+  }
+}
```
vara-bonthu, line 24: remove this. see the above comment

examples/analytics/emr-on-eks/main.tf
```diff
@@ -1,62 +1,40 @@
...
+
+  exec {
+    api_version = "client.authentication.k8s.io/v1alpha1"
+    command     = "aws"
+    # This requires the awscli to be installed locally where Terraform is executed
+    args = ["eks", "get-token", "--cluster-name", module.eks_blueprints.eks_cluster_id]
   }
-  host                   = data.aws_eks_cluster.cluster.endpoint
-  cluster_ca_certificate = base64decode(data.aws_eks_cluster.cluster.certificate_authority.0.data)
-  token                  = data.aws_eks_cluster_auth.cluster.token
```
vara-bonthu, line 32: I think we should keep this as data source unless we hit any specific issues.

modules/kubernetes-addons/agones/versions.tf
```diff
@@ -0,0 +1,14 @@
...
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 3.72"
+    }
+    kubernetes = {
+      source  = "hashicorp/kubernetes"
+      version = ">= 2.10"
+    }
+  }
+}
```
vara-bonthu, line 14: Do we need versions file for each sub folder add-on? We have this defined in Kuberentes root module. The same version will implicitly applies to all the sub folders.

vara-bonthu (commented):
LGTM! I have added few questions.

You also added few empty `outputs.tf` files for few add-ons. Is this to support pre-commit checks?

examples/analytics/emr-on-eks/main.tf
bryantbiggs: it is what is [recommended by the Kubernetes provider](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#exec-plugins) and also mirrors what is provided in a kubeconfig that is pulled from EKS. would there be any scenarios where we would recommend users NOT use the exec method?

deploy/pr/main.tf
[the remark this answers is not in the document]
bryantbiggs: do we still need the `vpc-test.yml` and `deploy/e2e/` as well?

modules/kubernetes-addons/agones/versions.tf
bryantbiggs: Your 2nd comment is correct, but because we are treating them as modules, then they hit these rules https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/.tflint.hcl#L15-L21

I would propose that we move these out of modules within the addon directory and this would reduce all this boilerplate

bryantbiggs:
regarding the versions and outputs - because we are stating that these are modules then the tflint rules apply; specifically:
- `terraform_standard_module_structure`
- `terraform_required_providers`
- `terraform_required_version`

if you copy the `.pre-commit-config.yaml` file from this PR and execute it locally, you can see the errors that show up from tflint

deploy/pr/versions.tf
bryantbiggs: removed in https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/424/commits/33053a8c1d5785a00c0859fbcb24827faad6475d

deploy/pr/main.tf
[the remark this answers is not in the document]
vara-bonthu: No we don't need that as well. We may need to cleanup the existing VPC by running `vpc-test-cleanup.yaml` workflow before removing the following files

`vpc-test.yaml` , `vpc-test-cleanup.yaml` `e2e-test.yaml` and `e2e-test-cleanup.yaml` . All these will be replaced with Terratest `e2e-terratest.yaml` 

`deploy/e2e/eks` and `deploy/e2e/vpc` can be removed which is used by the above workflows

modules/kubernetes-addons/agones/versions.tf
vara-bonthu: Ok.. so we need this to pass the tflint. I am fine with that.

examples/analytics/emr-on-eks/main.tf
vara-bonthu: To be honest i am not sure what problem that this can introduce by using **exec plugin**.  One of the customer i worked with was using this exec plugin approach and they didn't have permissions to execute aws cli. Didn't spend time to debug further instead I have asked them to change this to data source and that worked for them.

deploy/pr/main.tf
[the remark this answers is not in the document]
bryantbiggs: updated in https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/424/commits/4379d40215480a63baff31437b37ebb37b4645c9 - @Zvikan if you wouldn't mind taking a look at that commit to make sure everything aligns with your expectations

Zvikan (approved):

deploy/pr/main.tf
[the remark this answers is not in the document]
Zvikan: LGTM, thank you @bryantbiggs

vara-bonthu (approved):
LGTM! E2E tests complete?
It's worth creating a tag before merging this PR

PR #459, opened 2022-04-21
Comment out harcoded region names in windows node group example

naris-silpakit:
### What does this PR do?

This PR makes the example under `examples/node-group/windows-node-groups` compatible with different regions. It does this by commenting-out the filter statement in the availability zone data source that used hardcoded `us-west-2*` AZs. Otherwise this example fails out of the box in regions other than us-west-2.

The filter statement exists to show how to deal with the issue here: https://aws.amazon.com/premiumsupport/knowledge-center/eks-cluster-creation-errors/ so I opted to comment it out instead of removing it entirely so we don't lose that knowledge.

This PR also includes some formatting updates caught by running pre-commit.

### More

### For Moderators

### Additional Notes

Tested both in us-west-2 and GovCloud.

### DIFF
--- a/examples/node-groups/windows-node-groups/main.tf
+++ b/examples/node-groups/windows-node-groups/main.tf
@@ -31,10 +31,10 @@ data "aws_availability_zones" "available" {
   # Specify AZs to avoid EKS cluster creation error due to reduced capacity in an AZ.
   # Change the AZ names per capacity available in the region you selected.
   # https://aws.amazon.com/premiumsupport/knowledge-center/eks-cluster-creation-errors/
-  filter {
-    name   = "zone-name"
-    values = ["us-west-2a", "us-west-2b", "us-west-2c"]
-  }
+  # filter {
+  #   name   = "zone-name"
+  #   values = ["us-west-2a", "us-west-2b", "us-west-2c"]
+  # }
 }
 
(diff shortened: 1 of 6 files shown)


examples/node-groups/windows-node-groups/main.tf
```diff
@@ -31,10 +31,10 @@ data "aws_availability_zones" "available" {
   # Specify AZs to avoid EKS cluster creation error due to reduced capacity in an AZ.
   # Change the AZ names per capacity available in the region you selected.
   # https://aws.amazon.com/premiumsupport/knowledge-center/eks-cluster-creation-errors/
-  filter {
-    name   = "zone-name"
-    values = ["us-west-2a", "us-west-2b", "us-west-2c"]
-  }
+  # filter {
+  #   name   = "zone-name"
+  #   values = ["us-west-2a", "us-west-2b", "us-west-2c"]
```
Zvikan, line 36: I believe we want to keep this filter with the hard coded AZs, similar to our approach and upcoming PR https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/424 where we are keeping hard coded region by default, same stands true for this case.

naris-silpakit: Whoops, I missed that. I'll close this then.

PR #484, opened 2022-04-27
Examples to launch clusters with EC2 Spot Instances

christianhxc:
### What does this PR do?
Adds a set of examples for creating EKS clusters using EC2 Spot instances. The set of examples are for: managed node groups, self-managed node groups, and Karpenter. For self-managed node groups, the ASGs need to have the `mixed_instances_policy` block to override the list of instance types and configure Spot correctly. So, I had to modify the `aws-eks-self-managed-node-groups` module to support this. Additionally, this change in the module will allow to configure multiple instance types for on-demand as well.

### Motivation
Missing examples with EC2 Spot instances.

### More

### For Moderators

### Additional Notes

N/A


modules/aws-eks-self-managed-node-groups/main.tf
```diff
@@ -5,9 +5,45 @@ resource "aws_autoscaling_group" "self_managed_ng" {
   min_size            = local.self_managed_node_group["min_size"]
   vpc_zone_identifier = length(local.self_managed_node_group["subnet_ids"]) == 0 ? (local.self_managed_node_group["subnet_type"] == "public" ? var.context.public_subnet_ids : var.context.private_subnet_ids) : local.self_managed_node_group["subnet_ids"]
 
-  launch_template {
-    id      = module.launch_template_self_managed_ng.launch_template_id[local.lt_self_managed_group_map_key]
-    version = module.launch_template_self_managed_ng.launch_template_latest_version[local.lt_self_managed_group_map_key]
+  capacity_rebalance = local.self_managed_node_group["capacity_rebalance"]
+
+  dynamic "launch_template" {
```
bryantbiggs, line 10: this is now a breaking change so we'll have to figure out what our plan is - cc @askulkarni2 / @kcoleman731

bryantbiggs (dismissed):
Thanks for the PR!

Instead of creating new examples, can we instead utilize the respective existing examples and extend those to show spot usage?

For the EKS managed node groups and self managed node groups we can just add your node groups to those clusters.

For Karpenter we can just use `values: ["spot", "on-demand"]`

modules/aws-eks-self-managed-node-groups/main.tf
christianhxc: I'm happy to discuss this @bryantbiggs, the examples I uploaded are using only one instance type (as the rest of the examples in this repo) to test that path to make sure it's backwards compatible.

christianhxc:
@vara-bonthu I've moved the examples to the node-groups folder. I'll now work on the docs file to create a new section, but you should be able to start testing the spot examples 😀

examples/node-groups/self-managed-node-groups-spot/main.tf
```diff
@@ -0,0 +1,186 @@
...
+
+    spot_2vcpu_8mem = {
+      node_group_name    = "smng-spot-2vcpu-8mem"
+      capacity_type      = "spot"
+      capacity_rebalance = true
+      instance_types     = ["m5.large", "m4.large", "m6a.large", "m5a.large", "m5d.large"]
+      min_size           = "0"
+      subnet_ids         = module.aws_vpc.private_subnets
+      launch_template_os = "amazonlinux2eks" # amazonlinux2eks or bottlerocket
+      k8s_taints         = [{ key = "spotInstance", value = "true", effect = "NO_SCHEDULE" }]
```
vara-bonthu, line 140: Does this work? `k8s_taints` is not a valid input parameter for self-managed node groups.

Currently taints can be applied through `kubelet_extra_args` . I will update the docs for this

modules/aws-eks-self-managed-node-groups/locals.tf
```diff
@@ -62,13 +68,24 @@ locals {
...
   ]))
 
+  k8s_labels = merge(
+    local.self_managed_node_group["k8s_labels"],
+    {
+      "eks/capacityType"    = local.self_managed_node_group["capacity_type"],
+      "eks/node_group_name" = local.self_managed_node_group["node_group_name"]
+    }
+  )
+
```
vara-bonthu, line 78: Are(`k8s_labels`) these being applied anywhere in the code?

vara-bonthu (changes requested):
@christianhxc I am still testing your PR but i left few early comments for you.

examples/karpenter-with-spot/main.tf
```diff
@@ -0,0 +1,250 @@
...
+    helm = {
+      source  = "hashicorp/helm"
+      version = ">= 2.4.1"
+    }
+    kubectl = {
+      source  = "gavinbunney/kubectl"
+      version = ">= 1.13.1"
+    }
+  }
+
```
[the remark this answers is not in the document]
christianhxc: @vara-bonthu I get your point, let me work on those changes then :)

examples/node-groups/managed-node-groups-spot/main.tf
```diff
@@ -0,0 +1,179 @@
...
+
+  # EKS MANAGED NODE GROUPS
+  managed_node_groups = {
+    ondemand_2vcpu_8mem = {
+      node_group_name = "mng-ondemand-2vcpu-8mem"
+      instance_types  = ["m5.large", "m4.large", "m6a.large", "m5a.large", "m5d.large"]
+      min_size        = "2"
+      subnet_ids      = module.aws_vpc.private_subnets
+    }
+
```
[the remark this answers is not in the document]
christianhxc: @vara-bonthu sounds good, I'll work on this and if I see any challenge, I'll ping you back

examples/node-groups/self-managed-node-groups-spot/main.tf
```diff
@@ -0,0 +1,186 @@
+terraform {
+  required_version = ">= 1.0.1"
```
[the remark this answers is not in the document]
christianhxc: @vara-bonthu sounds good

examples/node-groups/self-managed-node-groups-spot/main.tf
christianhxc: @vara-bonthu yes, it works, but if the preference it's doing it through `kubelet_extra_args` I'll do it.

modules/aws-eks-self-managed-node-groups/locals.tf
christianhxc: @vara-bonthu I'm replicating what MGNs are doing. The `eks/capacityType` tag will be used to schedule pods in these nodes, and the `eks/node_group_name` is used to help CA identify which node groups to scale (we need this when the node groups wants to scale down to zero).

examples/karpenter-with-spot/main.tf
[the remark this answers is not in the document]
christianhxc: @vara-bonthu I've updated the Karpenter example and removed the specific one I created for Spot

examples/node-groups/managed-node-groups-spot/main.tf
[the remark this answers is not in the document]
christianhxc: @vara-bonthu I've updated the MNG example and removed the specific one I created for Spot

examples/node-groups/self-managed-node-groups-spot/main.tf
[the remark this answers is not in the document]
christianhxc: @vara-bonthu I've updated the SMNG example and removed the specific one I created for Spot

christianhxc:
@vara-bonthu I've fixed some problems that the automated tests reported, however there's one that's still failing due to some cache issues. Please give it a look and let me know. While you test the latest changes, I'm gong to work in updating the docs.

examples/node-groups/managed-node-groups/main.tf
```diff
@@ -260,6 +270,32 @@ module "eks_blueprints" {
         subnet_type = "private"
       }
     }
+    # Managed Node group with Launch templates using AMI TYPE and SPOT instances of 2 vCPUs and 8 Gib Memory
+    spot_2vcpu_8mem = {
+      node_group_name = "mng-spot-2vcpu-8mem"
+      capacity_type   = "SPOT"
+      instance_types  = ["m5.large", "m4.large", "m6a.large", "m5a.large", "m5d.large"]
```
bryantbiggs, line 277: we can't mix x86 and AMD instance types - the AMI pulled only supports one or the other

examples/node-groups/self-managed-node-groups/main.tf
```diff
@@ -27,6 +27,16 @@ locals {
     Blueprint  = local.name
     GithubRepo = "github.com/aws-ia/terraform-aws-eks-blueprints"
   }
+
+  ca_expander          = "priority"
```
bryantbiggs, line 31: same for this example, lets just put directly where used since its only one location

modules/aws-eks-managed-node-groups/locals.tf
```diff
@@ -107,15 +107,30 @@ locals {
...
+      Name                                                      = "${var.context.eks_cluster_id}-${local.managed_node_group["node_group_name"]}"
       "kubernetes.io/cluster/${var.context.eks_cluster_id}"     = "owned"
       "k8s.io/cluster-autoscaler/${var.context.eks_cluster_id}" = "owned"
       "k8s.io/cluster-autoscaler/enabled"                       = "TRUE"
   })
+
+  # NOTE: To support node groups to scale down to zero nodes,
+  # cluster autoscaler needs to more details to identify which nodes to scale-down.
+  # Here's the open issue and recommendations in GitHub: https://github.com/aws/containers-roadmap/issues/724
+  asg_tags = merge(
```
bryantbiggs, line 130: instead of baking this in here, we should move it out to the example. the way its written currently means its forcing these tags on users, even if they don't use cluster autoscaler

modules/aws-eks-managed-node-groups/main.tf
```diff
@@ -82,3 +82,17 @@ resource "aws_eks_node_group" "managed_ng" {
   ]
 
 }
+
+# NOTE: Tags need to replicated manually to the ASGs as per this issue: https://github.com/aws/containers-roadmap/issues/608
+# We won't need to have this resource when managed node groups in EKS replicate tags to ASGs
+resource "aws_autoscaling_group_tag" "cluster_autoscaler_tag" {
+  for_each = local.asg_tags
+
+  autoscaling_group_name = aws_eks_node_group.managed_ng.resources[0].autoscaling_groups[0].name
```
bryantbiggs, line 91: this can be problematic, we should avoid embedding this into the module - see https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1558#issuecomment-1030640207

modules/aws-eks-self-managed-node-groups/self-managed-launch-templates.tf
```diff
@@ -8,7 +8,7 @@ module "launch_template_self_managed_ng" {
       launch_template_os     = local.self_managed_node_group["launch_template_os"]
       launch_template_prefix = local.self_managed_node_group["node_group_name"]
       instance_type          = local.self_managed_node_group["instance_type"]
-      capacity_type          = local.self_managed_node_group["capacity_type"]
+      capacity_type          = local.self_managed_node_group["capacity_type"] == "spot" ? "" : local.self_managed_node_group["capacity_type"]
```
bryantbiggs, line 11: why do we need to make this change - shouldn't it work as its written currently?

modules/aws-eks-managed-node-groups/main.tf
christianhxc: @bryantbiggs what's the alternative then? I was able to create the cluster without any problems. The dependency issue was when I was trying to add a tag with the cluster id which you get it after the cluster is created. But the tags I left in the example work. To avoid this problem, this section could be **only** for the tags I'd use for CA to scale-down to zero nodes, and not leave it open to include any other tags. What do you think?

bryantbiggs: the guidance for tags on EKS managed node groups is:
1. Use a custom launch template which will propagate tags down to the ASGs and instances
2. If using default launch template provided by EKS managed node group, then users can use this `aws_autoscaling_group_tag` resource in their implementation

modules/aws-eks-self-managed-node-groups/self-managed-launch-templates.tf
christianhxc: when you're creating a spot ASG with mixed instance policies you need to leave it blank

modules/aws-eks-managed-node-groups/main.tf
bryantbiggs: > when we fix the tagging problem this hack will go away
Correct - which is why its better to externalize it. If we bake it into the module and remove it later, its a breaking change. If users use the tag resource and native support is offered through the API in the future, its a quality of life enhancement

examples/node-groups/managed-node-groups/main.tf
christianhxc: so, no change here, right? there's no ARM instance type in the list

modules/aws-eks-managed-node-groups/main.tf
christianhxc: so perhaps it makes sense to add another example that illustrates both scenarios? otherwise, this will make it existing examples way more complex ... I have customers asking me for an example on this one

modules/aws-eks-self-managed-node-groups/locals.tf
```diff
@@ -37,8 +40,11 @@ locals {
     subnet_ids              = []
     additional_tags         = {}
     additional_iam_policies = []
+    k8s_labels              = {}
```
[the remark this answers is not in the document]
vara-bonthu: I don't think `k8s_labels` used in self-managed. Labels and Taints should be passed through `kubelet_extra_args ` for self managed.

@christianhxc This line can be removed

examples/node-groups/managed-node-groups/main.tf
```diff
@@ -272,6 +308,18 @@ module "eks_blueprints_kubernetes_addons" {
 
   enable_metrics_server     = true
   enable_cluster_autoscaler = true
+  cluster_autoscaler_helm_config = {
+    set = [
+      {
+        name  = "extraArgs.expander"
+        value = local.ca_expander
```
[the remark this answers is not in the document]
christianhxc: done

examples/node-groups/self-managed-node-groups/main.tf
christianhxc: done

examples/node-groups/self-managed-node-groups/main.tf
```diff
@@ -120,6 +130,58 @@ module "eks_blueprints" {
...
+  eks_cluster_id           = module.eks_blueprints.eks_cluster_id
+  auto_scaling_group_names = module.eks_blueprints.self_managed_node_group_autoscaling_groups
+
+  # EKS Managed Add-ons
+  enable_amazon_eks_vpc_cni    = true
+  enable_amazon_eks_coredns    = true
+  enable_amazon_eks_kube_proxy = true
+
+  #K8s Add-ons
+  enable_aws_load_balancer_controller = true
```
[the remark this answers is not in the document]
christianhxc: done

modules/aws-eks-managed-node-groups/locals.tf
christianhxc: removed

modules/aws-eks-self-managed-node-groups/locals.tf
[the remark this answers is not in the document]
christianhxc: done

askulkarni2 (commented):
PR looks good to me. This is a breaking change however as @bryantbiggs called out and we need to address what's the best way to move forward.

askulkarni2 (approved):

docs/node-groups.md
```diff
@@ -460,6 +406,83 @@ The below example demonstrates advanced configuration options using Spot/GPU ins
...
+      create_launch_template = true              # false will use the default launch template
+      launch_template_os     = "amazonlinux2eks" # amazonlinux2eks or bottlerocket
+
+      # This is so cluster autoscaler can identify which node (using ASGs tags) to scale-down to zero nodes
+      additional_tags = {
+        "k8s.io/cluster-autoscaler/node-template/label/eks.amazonaws.com/capacityType" = "SPOT"
+        "k8s.io/cluster-autoscaler/node-template/label/eks/node_group_name"            = "mng-spot-2vcpu-8mem"
+      }
+    }
+```
```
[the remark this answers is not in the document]
FernandoMiguel: > Is it possible to `terraform state mv ...` this resource? If so maybe we can add instructions for that. 

Even better.... There is a terraform resource called moved that handles state operations in code.

https://www.terraform.io/language/modules/develop/refactoring

PR #492, opened 2022-04-30
chore: Update CI action workflows

bryantbiggs:
### What does this PR do?
- Re-enable `pre-commit.yaml` workflow which will allow users to run the same CI checks locally 
- Add plugin caching to the `pre-commit.yaml` workflow to speed up terraform validate checks which perform a `terraform init` on each project root in order to evaluate
- Remove workflows for kics, tfsec, and terrascan - a pre-commit check for tfsec is added to the pre-commit config to execute against the `examples/` directory which is the actual implementation and not the module definition. We can re-evaluate later what security tools are required to avoid redundancy and what areas of the codebase they should be evaluating; for now we'll keep it simple
- Remove `super-lint` workflow since its not used currently plus it appears to take quite some time with a lot of verbose results
- Update workflows to use Terraform 1.0.0 since that is the minimum supported version the project requires
- Add automatic example directory discovery to the tfplan workflow. Instead of hardcoding a list of example directories to evaluate, now the workflow will automatically discover any new examples added or if the directories are re-organized within the examples directory *provided there is a `versions.tf` file

### Motivation
- Improve the CI experience to increase time to feedback

### More

### For Moderators


bryantbiggs:
Looks like its catching failures with examples that weren't previously in the list of hardcoded examples so thats good news

.github/workflows/e2e-parallel-destroy.yml
```diff
@@ -25,48 +25,48 @@ jobs:
...
           - example_path: examples/fully-private-eks-cluster
-            tenant_name: "private"
+            tenant_name: 'private'
             deployment_order:
               [
-                "module.e2e_test.module.aws_vpc",
-                "module.e2e_test.module.vpc_endpoint_gateway",
-                "module.e2e_test.module.vpc_endpoints",
-                "module.e2e_test.module.eks_blueprints",
+                'module.e2e_test.module.aws_vpc',
```
bryantbiggs, line 43: if every `example/*` is structured the same, we could parameterize a lot of this instead of hardcoding the example names to evaluate

.github/workflows/e2e-parallel-destroy.yml
```diff
@@ -86,7 +86,6 @@ jobs:
             sed -i "s!TF_STATE_PATH!${{ matrix.example_path }}!g" deploy/e2e-test/backend.conf
             sed -i "s!EXAMPLE_PATH!${{ matrix.example_path }}!g" deploy/e2e-test/main.tf
           fi
-        continue-on-error: false
```
bryantbiggs, line 89: this is the default so we shouldn't need to add

.github/workflows/e2e-parallel-destroy.yml
```diff
@@ -113,30 +111,12 @@ jobs:
       - name: Terraform Job
         uses: hashicorp/setup-terraform@v1
         with:
-          terraform_version: 1.0.11
-
-      - name: Terraform Fmt
```
bryantbiggs, line 118: Because this is the destroy workflow, we shouldn't need to format, validate, plan - just tear it all down

.github/workflows/examples-tfplan-tests.yml
```diff
@@ -18,24 +34,7 @@ jobs:
...
-            examples/game-tech/agones-game-controller,
-            examples/ingress-controllers/nginx,
-            examples/karpenter,
-            examples/multi-tenancy-with-teams,
-            examples/node-groups/managed-node-groups,
-            examples/node-groups/self-managed-node-groups,
-            examples/node-groups/windows-node-groups,
-            examples/node-groups/fargate-profiles
-          ]
+        tf_working_dir: ${{ fromJson(needs.getExampleDirectories.outputs.directories) }}
```
bryantbiggs, line 37: validate #allTheThings 

For the ones we know won't pass (i.e. - external DNS where we don't have a proper domain to use yet), we can add a filter/exclusion for just those examples

.github/workflows/pre-commit.yaml
```diff
@@ -33,6 +34,16 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - name: Config Terraform plugin cache
```
bryantbiggs, line 37: For each "root" directory (i.e. - terraform default workspace location), our Terraform validate will do a `terraform init` which pulls down all the necessary providers, modules, etc. and that takes a LOOOOOOOOOOOONG time. I haven't used this before but I am hoping by centrally caching the plugin directory, that we can speedup that process and shave off some time

bryantbiggs:
Looks like `tfsec` is working now as well - https://github.com/aws-ia/terraform-aws-eks-blueprints/runs/6240218379?check_suite_focus=true

.pre-commit-config.yaml
```diff
@@ -33,3 +33,5 @@ repos:
           - '--args=--only=terraform_required_providers'
           - '--args=--only=terraform_standard_module_structure'
           - '--args=--only=terraform_workspace_remote'
+      - id: terraform_tfsec
+        files: ^examples/ # only scan `examples/*` which are the implementation
```
antonbabenko, line 37: 👍 

Some tools evaluate the code of the modules instead of the implementations and blame us for bad code... lol

.github/workflows/plan-examples.yml
```diff
@@ -0,0 +1,68 @@
...
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  getExampleDirectories:
+    name: Get example directories
+    runs-on: ubuntu-latest
+    # Skip running on forks since it won't have access to secrets
+    if: github.repository_owner == 'aws-ia'
```
bryantbiggs, line 14: Instead of showing failures to users on forks, we can just skip the checks until we develop a new strategy for executing plan validations on all PRs (forks included)

examples/eks-cluster-with-import-vpc/eks/main.tf
```diff
@@ -1,121 +0,0 @@
...
-    command     = "aws"
-    # This requires the awscli to be installed locally where Terraform is executed
-    args = ["eks", "get-token", "--cluster-name", module.eks_blueprints.eks_cluster_id]
-  }
-}
-
-#---------------------------------------------------------------
-# Terraform VPC remote state import from S3
-#---------------------------------------------------------------
-data "terraform_remote_state" "vpc_s3_backend" {
```
bryantbiggs, line 48: Technically this is just a data source and not an import whereas `terraform import <aws_resource> <identifier>` is what is referred to as an import

.github/workflows/e2e-parallel-destroy.yml
Zvikan: if `deployment_order` used, it usually means there's a custom order of modules&/resources which is not same order as `DEFAULT_DEPLOY_ORDER` which is what most examples use

.github/workflows/plan-examples.py
```diff
@@ -0,0 +1,27 @@
+import json
+import glob
+import re
+
+
+def get_examples():
+    """
+    Get all Terraform example root directories using their respective `versions.tf`;
+    returning a string formatted json array of the example directories minus those that are excluded
```
askulkarni2, line 9: We need to document that `versions.tf` is required for examples.

.github/workflows/plan-examples.yml
```diff
@@ -0,0 +1,68 @@
...
+
+    # These permissions are needed to interact with GitHub's OIDC Token endpoint.
+    permissions:
+      id-token: write
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        tf_working_dir: ${{ fromJson(needs.getExampleDirectories.outputs.directories) }}
```
askulkarni2, line 41: nice!

askulkarni2 (commented):
Looking good, just some minor clarifications.

.github/workflows/e2e-parallel-full.yml
```diff
@@ -85,7 +85,6 @@ jobs:
             sed -i "s!TF_STATE_PATH!${{ matrix.example_path }}!g" deploy/e2e-test/backend.conf
             sed -i "s!EXAMPLE_PATH!${{ matrix.example_path }}!g" deploy/e2e-test/main.tf
           fi
-        continue-on-error: false
```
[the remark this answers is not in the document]
Zvikan: false is default behavior

.github/workflows/e2e-parallel-full.yml
```diff
@@ -112,30 +110,26 @@ jobs:
       - name: Terraform Job
         uses: hashicorp/setup-terraform@v1
         with:
-          terraform_version: 1.0.11
+          terraform_version: 1.0.0
```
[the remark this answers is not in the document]
bryantbiggs: If we have to pick a version, my suggestion would be to use the minimum supported version https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/deec7d5caea47c06dea48fa616ad2c56e52c3cce/versions.tf#L2

.github/workflows/plan-examples.py
bryantbiggs: sure! What would be sufficient to cover this and where would it go?

Note: this logic is also the same logic used in https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/deec7d5caea47c06dea48fa616ad2c56e52c3cce/.github/workflows/pre-commit.yaml#L23

.github/workflows/pre-commit.yaml
```diff
@@ -66,6 +77,19 @@ jobs:
...
+        run: mkdir --parents ${{ env.TERRAFORM_DOCS_VERSION }}
+
+      - name: Cache Terraform
+        uses: actions/cache@v3
+        with:
+          path: ${{ env.TERRAFORM_DOCS_VERSION }}
+          key: ${{ runner.os }}-terraform-${{ hashFiles('**/.terraform.lock.hcl') }}
+          restore-keys: ${{ runner.os }}-terraform-
```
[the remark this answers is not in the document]
bryantbiggs: Correct - two separate jobs so two separate cache setups

Zvikan (approved):

Issue #510, opened 2022-05-05
[Bug]: Description is incorrect (copy/paste error) in `/outputs.tf` for `eks_cluster_status`


spkane:
### Welcome to Amazon EKS Blueprints!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/aws-ia/terraform-aws-eks-blueprints/issues) and didn't find any.

### Amazon EKS Blueprints Release version

f50247cb791782745315d9073509f8c5d0928dff

### What is your environment, configuration and the example used?

https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/outputs.tf#L40

### What did you do and What did you see instead?

It just needs a minor fix when someone has the time

PR #517, opened 2022-05-06
Add preserve for EKS Managed Addons

vara-bonthu:
### What does this PR do?

- Currently EKS Add-ons are controlled by the `kubernetes-addon`  sub module. VPC CNI add-on is getting deleted when we destroy kubernetes-addon before the cluster and node groups. 
- This feature will cleanup the EKS managed add-ons as and preserve the add-on as self-managed addon. This will ensure cirtical add-ons are running even after the add-ons and node groups are deleted
- Preserved self-managed addons will be deleted automatically when you delete the cluster
- This is added only for critical addons vpc-cni, coredns and kube-proxy. We will revise if its required for other managed add-ons
- If this users wanted to delete the add-on completely then can always pass `preserve=false`
- #510 - fixes the description issue

### More

### For Moderators

### DIFF
--- a/docs/add-ons/managed-add-ons.md
+++ b/docs/add-ons/managed-add-ons.md
@@ -71,7 +71,7 @@ Checkout this [doc](https://docs.aws.amazon.com/eks/latest/userguide/managing-vp
     addon_version            = "v1.4.0-eksbuild.preview"
     service_account          = "ebs-csi-controller-sa"
     resolve_conflicts        = "OVERWRITE"
-    namespace                = "kube-system"       
+    namespace                = "kube-system"
     additional_iam_policies  = []
     service_account_role_arn = ""
     tags                     = {}
--- a/docs/add-ons/managed-add-ons.md
+++ b/docs/add-ons/managed-add-ons.md
@@ -23,8 +23,9 @@ EKS managed add-ons can be enabled via the following.
     service_account          = "aws-node"
     resolve_conflicts        = "OVERWRITE"
     namespace                = "kube-system"
-    additional_iam_policies  = []
     service_account_role_arn = ""
+    preserve                 = true # This option makes the add-on a self-managed add-on, rather than an Amazon EKS add-on. There is no downtime for the add-on.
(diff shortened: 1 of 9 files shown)


docs/add-ons/managed-add-ons.md
```diff
@@ -37,6 +38,7 @@ EKS managed add-ons can be enabled via the following.
     resolve_conflicts        = "OVERWRITE"
     namespace                = "kube-system"
     service_account_role_arn = ""
+    preserve                 = true # This option makes the add-on a self-managed add-on, rather than an Amazon EKS add-on. There is no downtime for the add-on.
```
kcoleman731, line 41: since this is the managed addon example, we should prob show these as false in the docs. Also, can you expand on what you mean by no downtime?

modules/kubernetes-addons/aws-coredns/locals.tf
```diff
@@ -8,6 +8,7 @@ locals {
     namespace                = "kube-system"
     service_account_role_arn = ""
     additional_iam_policies  = []
+    preserve                 = true
```
kcoleman731, line 11: shouldn't this be false by default?

kcoleman731 (commented):
Comments for ya.

docs/add-ons/managed-add-ons.md
askulkarni2: Also should we call this something like 'enable_self_manage' or something? Just to make it super clear what this actually does. We can add more info in the docs.

vara-bonthu: @kenerwin88 Updated the PR with NOTE
@askulkarni2 This is actually an EKS Add-on built in option which we aren't leveraging so far. By default this option option is false. I have made it true for critical add-ons.

bryantbiggs (approved):
a variable for `preserve` would be good at some point (doesn't need to be now) because this is a common route for users to programmatically delete the managed addons

askulkarni2 (approved):
LGTM!

PR #530, opened 2022-05-11
fix: Remove default managed addon versions

bryantbiggs:
### What does this PR do?
- Removes the default addon versions for EKS managed addons. This is an [optional attribute](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_addon#addon_version) and when not specified, the EKS service will select the appropriate version to use for the clusters Kubernetes version
- Update `examples/complete-kubernetes-addons/*` example to fix errors:
	- Removed spark-operator and yunikorn addons since they are demonstrated in https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/examples/analytics/spark-k8s-operator
	- Removed EMR since it is demonstrated in https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/examples/analytics/emr-on-eks
	- Replaced EKS managed addon versions with addon version data source to demonstrate how users can select the default version or latest version of the addon
	- Added security group rules to open broad access for the various addons used until specific ports/protocols are identified
	- Removed Prometheus custom values - unable to identify what in these custom values is causing prometheus-server to fail, but works successfully without
- Correct `Preserve_Key` and `Reserve_Data` settings from string of `On` to boolean thanks to @Zvikan 's [findings](https://docs.fluentbit.io/manual/pipeline/filters/parser)
- EKS managed addons were updated to:
	- Remove default versions set
	- Hardcode addon name; this is not configurable by  users
	- Collapse use of `locals` down to `try()`
	- With the prior introduction of `preserve` attribute, the minimum supported provider version is updated to [v4.10.0](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md#4100-april-14-2022)
	- ebs-csi driver and vpc-cni addons were updated to exclude IAM resource creation when an IAM role is provided for use
	
### Motivation
- Closes #519
- Fixes the complete addon example to where the cluster and addons come up successfully
- Fixes came up while working towards resolving cluster upgrade - further work is required on this
- Updating the node-termination-handler version in the example fixed its crashloopbackoff but further investigation required. Relates to #507

### More

- [x] Yes, I have tested the PR using my local account setup  (Provide any test evidence report under Additional Notes)
- [ ] Yes, I have added a new example under [examples](https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/examples) to support my PR
[...]

PR #531, opened 2022-05-11
fix: Add ignore rules for tfsec and correct remaining issues to ensure tfsec passes

bryantbiggs:
### What does this PR do?
- Update tfsec to include ignore rules for common errors due to example usage. Notes have been added next to the rules for a brief explanation as to why in the `tfsec.yaml` file
- Remove `*.tfvars` and instead rely on default variables to avoid warnings on tfsec
- Fix various nits/issues raised by tfsec (security group descriptions, see individual ignore rules added within `*.tf` files
- Bastion EC2 removed from AMG OpenSearch example. Looks like it was a local testing artifact but no path to pem file - if we need to add this back in we can add an EC2 with SSM Session Manager access instead of public facing instance and pem key

### Motivation
- Fix tfsec failed errors on CI

### More

### For Moderators


examples/observability/eks-cluster-with-amp-amg-opensearch/main.tf
```diff
@@ -165,6 +165,7 @@ resource "grafana_data_source" "prometheus" {
 #---------------------------------------------------------------
 # Provision OpenSearch and Allow Access
 #---------------------------------------------------------------
+#tfsec:ignore:aws-elastic-search-enable-domain-logging
```
bryantbiggs, line 168: should be safe to ignore in example - otherwise requires creating CloudWatch log group in example

examples/observability/eks-cluster-with-amp-amg-opensearch/main.tf
```diff
@@ -234,15 +235,8 @@ resource "aws_elasticsearch_domain_policy" "opensearch_access_policy" {
 }
 
 resource "aws_security_group" "opensearch_access" {
-  vpc_id = module.aws_vpc.vpc_id
-
-  ingress {
```
bryantbiggs, line 239: was for the Bastion route - no longer a recommended practice to use this route but SSM Session Manger instead if needed

modules/aws-eks-managed-node-groups/managed-launch-templates.tf
```diff
@@ -34,9 +34,9 @@ resource "aws_launch_template" "managed_node_groups" {
   }
 
   metadata_options {
-    http_endpoint               = var.context.http_endpoint
-    http_tokens                 = var.context.http_tokens
-    http_put_response_hop_limit = var.context.http_put_response_hop_limit
+    http_endpoint               = try(var.context.http_endpoint, "enabled")
+    http_tokens                 = try(var.context.http_tokens, "required") #tfsec:ignore:aws-autoscaling-enforce-http-token-imds
```
bryantbiggs, line 38: tfsec wants `http_tokens` to be `required` but I don't know the consequences of enforcing that on all users. Instead, adding the `try()` block to default to a sane security setting for the IMDS and an ignore rule to silence the error

modules/kubernetes-addons/agones/main.tf
```diff
@@ -19,11 +19,12 @@ resource "kubernetes_namespace_v1" "this" {
...
 
 resource "aws_security_group_rule" "agones_sg_ingress_rule" {
+  description       = "Allow UDP ingress from internet"
   type              = "ingress"
   from_port         = local.helm_config["gameserver_minport"]
   to_port           = local.helm_config["gameserver_maxport"]
   protocol          = "udp"
-  cidr_blocks       = ["0.0.0.0/0"]
-  ipv6_cidr_blocks  = ["::/0"]
+  cidr_blocks       = ["0.0.0.0/0"] #tfsec:ignore:aws-vpc-no-public-ingress-sgr
```
bryantbiggs, line 27: Should we keep these like this - is it required or can we add variables and let users pick, but in the examples we can use these values?

modules/kubernetes-addons/aws-node-termination-handler/main.tf
```diff
@@ -29,6 +29,7 @@ resource "aws_autoscaling_group_tag" "aws_node_termination_handler_tag" {
   }
 }
 
+#tfsec:ignore:aws-sqs-enable-queue-encryption
```
bryantbiggs, line 32: SSE encryption is enabled, but looks like tfsec wants CMK encryption so silencing error for now

bryantbiggs:
pre-commit will fail on tflint - this is already fixed in my other PR https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/530/commits/1f61842e7bd801b962f954afb01c5dfbe68c17c4 so not re-adding here for duplication

modules/aws-eks-managed-node-groups/managed-launch-templates.tf
askulkarni2, line 38: Is there a way we could not ignore this one?

bryantbiggs: yes, it would be changing line 38 to 
```hcl
    http_tokens                 = "required"
```

However, I don't know the ramifications of this change on users and their applications

bryantbiggs: For reference - https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1749

kcoleman731 (approved):
This all looks good to me.

Zvikan (approved):

askulkarni2 (approved):
LGTM

PR #544, opened 2022-05-12
fix: Remove tfvars from terratest and update managed addons docs

bryantbiggs:
### What does this PR do?
- Removes `tfvars` setting from terratest - the tfvars file was previously removed and the provide variables default values are used instead
- Update cluster upgrade docs for upgrading addons 

### Motivation
- Fix e2e test
- Align docs with current implementation

### More

### For Moderators


askulkarni2 (approved):

bryantbiggs:
looks like I need to add a default value for the observability stuff - one sec

test/src/eks_blueprints_e2e_test.go
```diff
@@ -120,7 +120,7 @@ func TestEksBlueprintsE2E(t *testing.T) {
 			inputTfOptions := &terraform.Options{
 				/*The path to where our Terraform code is located*/
 				TerraformDir: tempExampleFolder,
-				VarFiles:     []string{testCase.name + ".tfvars"}, // The var file paths to pass to Terraform commands using -var-file option.
+				// VarFiles:     []string{testCase.name + ".tfvars"}, // The var file paths to pass to Terraform commands using -var-file option.
```
vara-bonthu, line 123: if we are not using the `tfvars` then we need align the cluster name through variables. Currently tfvars using the cluster name as `aws-terra-test-eks` and we are using this to validate the terra tests.

Updating the `variables.tf` should fix the issue `tenant=aws`, `environment=terra` and `zone=test`

vara-bonthu (changes requested):
LGTM! Change required for Terratest

test/src/eks_blueprints_e2e_test.go
bryantbiggs: thanks for the catch - I updated terratest to set the variables to match the expected cluster ID as you've shown above. I figure this is probably safer in case both terratest and the e2e-parallel are deployed at the same time (avoid name conflicts). Fixed in https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/544/commits/e32e19a332a0d984e7c696607736624b8109f3ad - let me know what you think

bryantbiggs: And https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/544/commits/289acf5080d9fd95a22a899da80c42924485c349 - forgot it was in two places

vara-bonthu: Thanks @bryantbiggs! This looks great 👍🏼

vara-bonthu (approved):
LGTM

PR #553, opened 2022-05-13
Fix/507 fix nth

### LINKED ISSUE #507: [QUESTION] aws-node-termination-handler SQS Queue
#### Please describe your quesiton here
aws-node-termination-handler module creates SQS resource as well as cloudwatch event rules resources, and by default configured to work in Queue mode (it has `enableSqsTerminationDraining: true` in values.yaml). What is the reason of this? 
I mean by default after creating it throws a lot of errors:
`2022/05/04 16:25:15 WRN There was a problem monitoring for events error="AWS.SimpleQueueService.NonExistentQueue: The specified queue does not exist for this wsdl version.\n\tstatus code: 400, request id: e7866f99-607d-579d-bd47-3575cb56d5fb" event_type=SQS_TERMINATE`
So basically it doesn't work as expected. Why not to configure queueURL variable in values.yaml?
```
# queueURL Listens for messages on the specified SQS queue URL
# queueURL: ${queueURL}
```
Or maybe I'm missing something and there is a way to configure that?

askulkarni2:
### What does this PR do?

AWS Node Termination Handler will now use IMDS processor by default. Queue processor can be configured by setting `aws_node_termination_handler_queue_processor = true`.

Fixes #507 

### More

- [X] Yes, I have added a new example under [examples](https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/examples) to support my PR

### For Moderators


bryantbiggs:
I think this looks great but my one, big question would be - can we achieve the same outcome without the use of sub-modules for queue based vs IMDS? It looks like (and correct me if I am wrong) the module usage here is so that you can clearly differentiate between using the queue based vs IMDS. However, we can still flatten that same logic into the root NTH addon module. Why flatten vs nesting in sub-modules:
1. We aren't re-using these modules anywhere nor are we looking to offer them as standalone usage modules; they are intended to serve the NTH addon module only
2. It adds a bit of overhead that adds up over time. Theres the boilerplate of all the now triplicated `variables.tf`, `README.md`, etc. And if you want to know why `terraform validate` is so slow, run it on an example that uses this (or any addons really), and then inspect the contents of the `.terraform/modules/*`

PR #555, opened 2022-05-15
Node groups updates

### LINKED ISSUE #468: [Bug]: Disk provisioning logic fails for managed node groups
### Welcome to Amazon EKS Blueprints!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/aws-ia/terraform-aws-eks-blueprints/issues) and didn't find any.

### Amazon EKS Blueprints Release version

v4.0.3 and main

### What is your environment, configuration and the example used?

Using the example from https://github.com/ondat/terraform-eksblueprints-ondat-addon/blob/main/blueprints/getting-started/main.tf, removing the ref pins to f22c58.

### What did you do and What did you see instead?

A `terraform apply` resulted in a timeout as the nodes never joined the cluster. Further investigation of the EKS managed node group nodes revealed that cloud-init failed due to:

```shell
+ IDX=1
++ awk '/disk/ {print $1}'
++ lsblk -o NAME,TYPE -dsn
+ DEVICES='nvme1n1
└─nvme0n1
nvme2n1
nvme3n1'
+ for DEV in $DEVICES
+ mkfs.xfs /dev/nvme1n1
meta-data=/dev/nvme1n1           isize=512    agcount=16, agsize=1638400 blks
         =                       sectsz=512   attr=2, projid32bit=1
         =                       crc=1        finobt=1, sparse=1, rmapbt=0
         =                       reflink=1
data     =                       bsize=4096   blocks=26214400, imaxpct=25
         =                       sunit=1      swidth=1 blks
naming   =version 2              bsize=4096   ascii-ci=0, ftype=1
log      =internal log           bsize=4096   blocks=12800, version=2
         =                       sectsz=512   sunit=1 blks, lazy-count=1
realtime =none                   extsz=4096   blocks=0, rtextents=0
+ mkdir -p /local1
+ echo /dev/nvme1n1 /local1 xfs defaults,noatime 1 2
+ IDX=2
+ for DEV in $DEVICES
+ mkfs.xfs /dev/└─nvme0n1
Error accessing specified device /dev/└─nvme0n1: No such file or directory
Usage: mkfs.xfs
[snip]
```

The loop seems to have become confused as our disks are pre-provisioned and pre-partitioned by the `ebs-bootstrap` component so they already have a partition ready and are populated with data when the code in https://github.com/aws-ia/terraform-aws-eks-blueprints/commit/f22c583290b2321a2bb169f11b57e7bc9d83ccfa#diff-de4948d167bbfc9b217e569f535730d007e98ee2702514233db204160a5e893dR18 runs.

### Additional Information

```shell
This would be a serious issue if somebody attached disks with state as it could lead to lost data - the mkfs.xfs is done regardless of the existing content of the disks. It's not clear how to opt out of this logic.
```

vara-bonthu:
### What does this PR do?

- New parameter introduced for Node groups to format and mount NVMe disk. Its false by default. fixes #468 
- New feature - Custom IAM roles for EKS Cluster, Node groups(Managed and self managed) and Fargate
- Improvements to Managed node groups update config
- Fix for VPC CNI upgrade issue
- Fix for Managed node group cluster version when custom AMI is used
- Node groups Examples updated with user IAM roles, update config and self managed node taints and labels
- Docs updated with the above new options

### More

### For Moderators


docs/node-groups.md
```diff
@@ -156,12 +161,15 @@ The below example demonstrates advanced configuration options using Spot/GPU ins
...
+        max_unavailable            = 1
+      }]
 
       # Node Group compute configuration
-      ami_type       = "AL2_x86_64"
-      capacity_type  = "SPOT"
-      instance_types = ["t3.medium", "t3a.medium"]
+      ami_type               = "AL2_x86_64"
+      capacity_type          = "SPOT"
+      instance_types         = ["r5d.xlarge", "r5a.xlarge"]
```
bryantbiggs, line 171: So we don't throw off users since this requires the `AL2_ARM_64` AMI type
```suggestion
      instance_types         = ["r5d.xlarge", "r5.xlarge"]
```

modules/aws-eks-managed-node-groups/iam.tf
```diff
@@ -1,4 +1,6 @@
 resource "aws_iam_role" "managed_ng" {
+  count = local.create_iam_role_count
```
bryantbiggs, line 2: just FYI - this is now a breaking change

modules/aws-eks-managed-node-groups/templates/userdata-amazonlinux2eks.tpl
```diff
@@ -11,7 +11,8 @@ echo "Running custom user data script"
 # User-supplied pre userdata
 ${pre_userdata}
 
-# Format and Mount NVMe Disks if available
+if [ ${format_mount_nvme_disk} = true ];then
```
bryantbiggs, line 14: Question: should we have this baked into the module or should we just demonstrate how users can add this to `pre_userdata` through an example? Any changes here will be impacting to both users who utilize this functionality or not (until [this](https://github.com/hashicorp/terraform-provider-aws/pull/23604) is supported by launch templates)

modules/aws-eks-self-managed-node-groups/locals.tf
```diff
@@ -54,13 +59,13 @@ locals {
   ec2_principal     = "ec2.${var.context.aws_partition_dns_suffix}"
 
   # EKS Worker Managed Policies
-  eks_worker_policies = toset(concat([
+  eks_worker_policies = local.create_iam_role ? toset(concat([
```
bryantbiggs, line 62: Does this throw a type mismatch error? `toset({})` vs `[]`? You might have to do:
```hcl
eks_worker_policies = {for k,v in toset(concat([
    "${local.policy_arn_prefix}/AmazonEKSWorkerNodePolicy",
    "${local.policy_arn_prefix}/AmazonEKS_CNI_Policy",
    "${local.policy_arn_prefix}/AmazonEC2ContainerRegistryReadOnly",
    "${local.policy_arn_prefix}/AmazonSSMManagedInstanceCore"],
    local.self_managed_node_group["additional_iam_policies"
  ])) : k => v if local.create_iam_role }
```

docs/node-groups.md
vara-bonthu: In this case i am looking for instance type with NVMe disks to show the formatting. I will update this example with correct instance types

docs/node-groups.md
```diff
@@ -156,12 +161,15 @@ The below example demonstrates advanced configuration options using Spot/GPU ins
...
+      format_mount_nvme_disk = true # Ephemeral storage disks are formatted and mounted under `local1` folder
```
askulkarni2, line 172: nice!

modules/aws-eks-managed-node-groups/iam.tf
vara-bonthu: I am adding this for users who want to bring their own IAM role to node groups. May be we can advise on mv state file as you recommended. I think its better to do it now rather than later

modules/aws-eks-self-managed-node-groups/iam.tf
```diff
@@ -1,4 +1,6 @@
 resource "aws_iam_role" "self_managed_ng" {
+  count = local.create_iam_role_count
```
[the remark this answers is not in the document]
askulkarni2: Should this be in the release notes?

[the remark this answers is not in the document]
vara-bonthu: Same as above comment for managed node groups. May be we can add this recommendation to known_issues guide or somewhere in docs?

askulkarni2 (commented):
Looking good, would like to see some test results.

modules/aws-eks-managed-node-groups/templates/userdata-amazonlinux2eks.tpl
vara-bonthu: I thought about moving this to pre-userdata but we already have this block of code for both managed and self-managed launch templates. So, it's easier to add a condition with a new variable.

This change wont impact the existing node groups unless they upgrade their node group which creates a new LT with this change. We can make this  to `true` by default so that it wont impact even if the upgrade happens. ONDAT team may need to send `format_mount_nvme_disk=false` for their usecase.

bryantbiggs: Right, the impact of formatting or not formatting won't have an impact, but I'm talking more generically in terms of any changes (ANY changes) within user data means the hash is changed and there will be a new diff that launch templates will want to roll out. I haven't looked at what our default strategy is for launch template updates, but this is one of the reasons I fought very hard to keep any customziations out of the provided user data because it affects everyone (a whitespace change will affect everyone) and the more customizations there are, the more updates to the user data, and the more changes felt by users. Just something to consider if this is not expected to be used by 80%+ of blueprints users but only a select few

bryantbiggs (approved):

Issue #523, opened 2022-05-10
[FEATURE] Addons such as aws_vpc_cni should optionally be installable via helm


andrewhibbert:
#### Is your feature request related to a problem? Please describe

Since many people use custom networking for aws_vpc_cni the modules should allow this to either be configurable via add on or via helm so this extra config can be set.

#### Describe the solution you'd like

Option to install via helm and provide helm values

bryantbiggs:
Thank you for the request @andrewhibbert ! Would you mind sharing how you manage this in your cluster today? Do you manually remove the EKS provided VPC CNI plugin and then re-install using Helm? Is your primary use case here to be able to configure [VPC CNI custom networking](https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html) or do you have other use cases you can share?

andrewhibbert:
Currently we just edit the resources on the fly, but we'd like to do this via code. Primary use case is for [VPC CNI custom networking](https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html), though I think the module should just provide minimal helm config similar to things like the karpenter module

lupindeterd:
@bryantbiggs I'm not sure if this relevant to this but it seems that the custom value I set through amazon_eks_vpc_cni_config doesn't get applied to the running DaemonSet, I want to change the SNAT env to true from false.

` amazon_eks_vpc_cni_config = 
{
    set = [{
      name  = "env.AWS_VPC_K8S_CNI_EXTERNALSNAT"
      value = "true"
      }
    ]
  }
`

bryantbiggs:
> @bryantbiggs I'm not sure if this relevant to this but it seems that the custom value I set through amazon_eks_vpc_cni_config doesn't get applied to the running DaemonSet, I want to change the SNAT env to true from false.
> 
> `amazon_eks_vpc_cni_config = { set = [{ name = "env.AWS_VPC_K8S_CNI_EXTERNALSNAT" value = "true" } ] }`

@lupindeterd - the VPC CNI addon is a managed addon and therefore does not currently support custom configurations such as setting environment variables.

bryantbiggs:
with EKS now supporting addon custom configuration, we are going to close out this issue. thank you!

PR #551, opened 2022-05-13
refactor: Remove `aws-resource-tags` module and its usage from blueprints

bryantbiggs:
### What does this PR do?
1. Removes the sub-module `aws-resource-tags` and its usage through the blueprints/examples
2. Update examples to create three distinct sections
	- Provider & locals - mostly setup to execute the example
	- The blueprint definition (cluster and addons) 
	- Supporting resources - any resources that are not blueprint specific but required to make the example deployable. The resources contained in this section is assumed to be left up to users (create new like example, integrate with existing, etc.)
3. Various updates
	- For numeric values, use numbers instead of strings
	- Rename `aws_vpc` to just simply `vpc` 

### Motivation
1. Tags are typically very specific to a users environment and their current standards, practices, business requirements, etc. Instead of being overly prescriptive on tags within the blueprint definition, we can instead leave that up to users unless there is a strong reason that is directly related tied to blueprints. In addition, the tag module is also serving as a naming resource which again is enforcing a naming scheme upon users instead of letting users pick the naming scheme which can be problematic when conflicting with business/compliance requirements
	- Relates to #333 as well
2. Clarify how the information is presented to users to make it easier to see what portion is relevant to blueprints, and what is not (there to support deploying the example). This also makes it easier for users to copy+paste the blueprint section where only minimal changes are required to implement within their environment
3. Closes #543 

### More

- [ ] Yes, I have tested the PR using my local account setup  (Provide any test evidence report under Additional Notes)
- [ ] Yes, I have added a new example under [examples](https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/examples) to support my PR
- [ ] Yes, I have created another PR for add-ons under [add-ons](https://github.com/aws-samples/eks-blueprints-add-ons) repo (if applicable)
- [ ] Yes, I have updated the [docs](https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/docs) for this feature
- [x] Yes, I ran `pre-commit run -a` with this PR

**Note**: Not all the PRs required examples and docs except a new pattern or add-on added.

### For Moderators
- [ ] E2E Test successfully complete before merge?

PR #558, opened 2022-05-17
feat: Add support for self-managed CoreDNS addon and example of serverless cluster using Fargate Profiles

### LINKED ISSUE #549: [FEATURE] Blueprint for serverless Fargate cluster
#### Is your feature request related to a problem? Please describe
- A blueprint that demonstrates how to create a serverless based EKS cluster using Fargate profiles. This solution will rely on showing users how they need to manage patching CoreDNS to avoid having to run an EC2 based node for essential addons

#### Describe the solution you'd like
- Blueprint of EKS cluster with Fargate profiles for data plane

#### Describe alternatives you've considered
- None

bryantbiggs:
### What does this PR do?
- Adds support for self-managed CoreDNS addon
- Adds example blueprint for Fargate Serverless cluster (data plane) using the self-managed CoreDNS addon; demonstrating how to modify the EKS provided CoreDNS deployment, etc.
- Fixes pre-commit checks from prior PR

### Motivation
- Closes #549
- Provides users with an example path on how to provision serverless EKS data planes using Fargate Profiles
- Provides the start of how we might be able to manage the option to use either EKS managed addon or self-managed addon; this has come up before with requests like #523  

### More

	- ⚠️ No - I can add it to that repo but I think we need to test and validate that the GitOps route works first due to the tricky nature of how this is managed currently

### For Moderators


examples/fargate-serverless/main.tf
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  enable_amazon_eks_coredns = true
+  coredns_use_managed_addon = false
+  coredns_helm_config = {
+    values = [
+      # The annotations are required to ensure the Fargate provisioner is used and not the EC2 provisioner
+      <<-EOT
+      image:
+        repository: 602401143452.dkr.ecr.${local.region}.amazonaws.com/eks/coredns
```
kcoleman731, line 110: Should this live in the add-on implementation?

modules/aws-eks-fargate-profiles/main.tf
```diff
@@ -1,19 +1,19 @@
 resource "aws_eks_fargate_profile" "eks_fargate" {
   cluster_name           = var.context.eks_cluster_id
-  fargate_profile_name   = "${local.fargate_profiles["fargate_profile_name"]}-${var.context.eks_cluster_id}"
```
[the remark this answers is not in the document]
kcoleman731: @vara-bonthu any issues here?

examples/fargate-serverless/main.tf
bryantbiggs: Great question - I wasn't sure to be honest. The reason why I left it here is because the version is somewhat "hidden" and we're using the `aws_eks_addon_version` data source to pull the version. If we move this down to the addon:
1. How do we manage the version? The `aws_eks_addon_version` data source needs the cluster version to know what the appropriate version is
2. If we do use the `aws_eks_addon_version`, then we need to take in more details like cluster version, and use most recent version or default, etc.

The versioning on this addon is not as straighforward as a standard helm chart addon

bryantbiggs: Actually - this would compliment the managed addon that is there. Let me move that down to the addon - it should simplify version management for the managed addon for users as well

examples/fargate-serverless/main.tf
```diff
@@ -0,0 +1,224 @@
...
+  triggers = {}
+
+  provisioner "local-exec" {
+    interpreter = ["/bin/bash", "-c"]
+    environment = {
+      KUBECONFIG = base64encode(local.kubeconfig)
+    }
+
+    # We are removing the deployment provided by the EKS service and replacing it through the self-managed CoreDNS Helm addon
+    # However, we are maintaing the existing kube-dns service and annotating it for Helm to assume control
```
FernandoMiguel, line 185: can you explain this a little better?
what does helm assume control mean here?

modules/aws-eks-fargate-profiles/main.tf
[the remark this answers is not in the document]
bryantbiggs: In theory, users can set the same name scheme defined previously if this change were to go through. If today they set:
```hcl
  cluster_name = "bar"
  ...
  fargate_profile_name = "foo"
  ...
```

They can set the following to avoid replacement/disruption:
```hcl
  cluster_name = "bar"
  ...
  fargate_profile_name = "foo-bar"
  ...
```

FernandoMiguel:
my only problem with this example is that it assumes creds are available to the null-exec.
but nor does it not take a profile flag, nor it accepts cross-account assume roles.
I've already spent days trying to work that out without much luck :(

bryantbiggs:
> my only problem with this example is that it assumes creds are available to the null-exec. but nor does it not take a profile flag, nor it accepts cross-account assume roles. I've already spent days trying to work that out without much luck :(

It uses an ephemeral kubeconfig which is contructed at apply time using an [auth token](https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/558/files#diff-369b463e8d2cb8740ca4bca4d7966254f8c3b51f9016c14be3ca8375752f248cR169). This means that all authorization/authentication is left with the provider which is left up to the users discretion. The current provider implementation uses the provider recommended `exec` which does accept `--profile`

examples/fargate-serverless/main.tf
bryantbiggs: Updated in https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/558/commits/22401ca4fc5db75e6be35f863ad2c7a52d621dd8

modules/kubernetes-addons/locals.tf
```diff
@@ -44,30 +44,30 @@ locals {
   # For addons that pull images from a region-specific ECR container registry by default
   # for more information see: https://docs.aws.amazon.com/eks/latest/userguide/add-ons-images.html
   amazon_container_image_registry_uris = tomap({
-    af-south-1     = "877085696533.dkr.ecr.af-south-1.amazonaws.com/amazon",
```
bryantbiggs, line 47: the repository name `amazon` was moved to the ALB ingress controller so that the registry URL can be shared amongst AWS addons (ALB ingress controller and CoreDNS currently, but VPC CNI, kube-proxy, etc. can follow this as well in the future)

modules/aws-eks-fargate-profiles/main.tf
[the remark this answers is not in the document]
vara-bonthu: This was updated in my existing PR. Agree with the change

examples/fargate-serverless/README.md
```diff
@@ -0,0 +1,66 @@
+# Serverless EKS Cluster using Fargate Profiles
+
+This example shows how to provision a serverless cluster (serverless data plane) using Fargate Profiles.
+
+There are some challenges to achieve this, primarily around the management and usage of CoreDNS on Fargate. There are alternative solutions shown on the internet, typically utilizing a `kubectl patch ...` command to update the compute type of the template spec annotation. This approach does work, but with the downside that the EKS service will eventually update the CoreDNS deployment and overwrite the patch, causing the CoreDNS deployment to fail due to insufficient capacity (it will be looking for EC2 compute when only Fargate is available).
```
vara-bonthu, line 5: Should we move this example under `examples\node-groups`?

vara-bonthu, line 5: Should we move this example under `examples\node-groups`?

vara-bonthu (commented):
I like the example for Fargate Serverless. My main comment is around the Self managed add-on for CoreDNS. I feel it should be a dedicated helm add-on and customers should be able to send their own config through `values.yaml`.  We can delete the existing CoreDNS add-on(deployed by the EKS cluster) when customers enables the self-managed add-on seamlessly.

examples/fargate-serverless/main.tf
```diff
@@ -0,0 +1,232 @@
...
+  enable_amazon_eks_kube_proxy = true
+  amazon_eks_kube_proxy_config = {
+    addon_version     = data.aws_eks_addon_version.latest["kube-proxy"].version
+    resolve_conflicts = "OVERWRITE"
+  }
+
+  enable_amazon_eks_coredns = true
+  coredns_use_managed_addon = false
+  coredns_helm_config = {
+    # Sets the correct annotations to ensure the Fargate provisioner is used and not the EC2 provisioner
```
FernandoMiguel, line 106: what's the reason for coredns not be the lastest version by default?

examples/fargate-serverless/README.md
bryantbiggs: Hmm, I'm not sure. Whats your thoughts/motivation around moving it to that directory?

examples/fargate-serverless/main.tf
bryantbiggs: It matches the API by using the default version. Safer route for users

examples/fargate-serverless/main.tf
```diff
@@ -0,0 +1,232 @@
...
+# Supporting Resources
+#---------------------------------------------------------------
+module "vpc" {
+  source  = "terraform-aws-modules/vpc/aws"
+  version = "~> 3.0"
+
+  name = local.name
+  cidr = "10.0.0.0/16"
+
+  azs             = ["${local.region}a", "${local.region}b", "${local.region}c"]
```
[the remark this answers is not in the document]
bryantbiggs: The a, b, or c AZ wasn't available or the region didn't support 3 AZs? My thoughts were to reduce all the extra configurations for VPC since the focus is on blueprints and not networking. In https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/551 I have updated all examples to the format shown here

modules/kubernetes-addons/aws-coredns/main.tf
```diff
@@ -1,7 +1,20 @@
...
+
+data "aws_eks_addon_version" "this" {
+  addon_name = local.name
+  # Need to allow both config routes - for managed and self-managed configs
+  kubernetes_version = try(var.addon_config.kubernetes_version, var.helm_config.kubernetes_version)
+  most_recent        = try(var.addon_config.most_recent, var.helm_config.most_recent, false)
+}
+
 resource "aws_eks_addon" "coredns" {
+  count = var.use_managed_addon ? 1 : 0
```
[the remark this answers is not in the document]
bryantbiggs: Do you mean, use the existing variable name? We could, but I think something like `use_managed_addon` is more clear about its intention: use the managed addon (`true`) or the self-manged addon (`false`). The  `enable_amazon_eks_coredns` variable has a different meaning within the addons root module which cause confusion as well if its re-used here

modules/kubernetes-addons/aws-coredns/main.tf
```diff
@@ -11,3 +24,45 @@ resource "aws_eks_addon" "coredns" {
     try(var.addon_config.tags, {})
   )
 }
+
+module "helm_addon" {
+  source = "../helm-addon"
+  count  = var.use_managed_addon ? 0 : 1
```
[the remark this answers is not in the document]
bryantbiggs: From an addon perspective, they are fulfilling a similar purpose: CoreDNS. The difference is - is it managed by the EKS service or by the user via a helm chart. 

I would like to avoid the expansion of modules because they come with a lot of overhead and they aren't really living up to modules. But, I would love to hear more on why this is not a good approach

examples/fargate-serverless/main.tf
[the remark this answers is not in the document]
FernandoMiguel: we tend to use this
```
data "aws_availability_zones" "good_zone_ids" {
  exclude_zone_ids = [
    # There are no modern instance types in use1-az3; AWS has left
    # that zone behind for the most part.  See, for example, here:
    # https://www.reddit.com/r/aws/comments/g5lh7h/t3m5r5_instance_types_in_use1az3/
    "use1-az3",
  ]
  state = "available"
}
```

and a slice can be used, like in other parts of the blueprints code already show
`  azs = slice(data.aws_availability_zones.available.names, 0, 3)`

.github/workflows/plan-examples.py
```diff
@@ -9,10 +9,8 @@ def get_examples():
     returning a string formatted json array of the example directories minus those that are excluded
     """
     exclude = {
-        'examples/eks-cluster-with-external-dns',  # excluded since GitLab auth, backend, etc. required
-        'examples/ci-cd/gitlab-ci-cd',  # excluded until Rout53 is setup
-        'examples/observability/amp-amg-opensearch',  # TODO - needs to be updated to run example
```
bryantbiggs, line 14: These two can now be planned successfully, so removing from exclusion list

examples/eks-cluster-with-new-vpc/main.tf
```diff
@@ -49,7 +49,8 @@ locals {
 module "eks_blueprints" {
   source = "../.."
 
-  cluster_name    = local.name
+  #                 var.cluster_name is for Terratest
```
bryantbiggs, line 52: this should fix Terratest when merged to main

vara-bonthu (approved):
LGTM! It would be good if we update docs(github pages) to show to how to consume self-managed CoreDNS with custom `values.yaml`.

PR #578, opened 2022-05-25
[teams] Makes quotas and labels optionals

aleixrm-scopely:
### What does this PR do?

Makes the quotas and labels fields in the team definition object optional, as they should not be required in most of the cases.

### Motivation

When creating the namespace for a specific team, if the `application_teams.*.label` object input field is not set, the terraform apply fails as the `labels` input of the resource expects that field to exist.

The same happens with the defined `kubernetes_resource_quota.team_compute_quota` and `kubernetes_resource_quota.team_object_quota ` resources, as they are not prepared for supporting empty or partial defined `application_teams.*.quota` inputs, and if any of them doesn't exist, the kubernetes provider fails with a go panic error due to the `kubernetes_resource_quota`'s `spec` resource field emptiness.

The changes in this PR filter and create two different quota maps for being applied in the respective namespace classifying each team's kubernetes namespace quotas on compute quotas or object quotas.

Maybe there are more efficient or clearer ways for doing that maintaining back-compatibility with the existing quota definition int the application_teams input, any suggestion for improving this will be welcome 👍

### More

- [ ] Yes, I have tested the PR using my local account setup  (Provide any test evidence report under Additional Notes) -> I wasn't able to execute the repository tests locally as I don't have the adequate infrastructure available, I manually tested the changes over an already created cluster with this module.

### For Moderators

### Additional Notes

Manual testing inputs:
```
application_teams = {
    test-team = {
      users = [ //OBFUSCATED ]
    },
    test-team2 = {
      labels = {
        "test2" = "test2"
      }
      quota = {
        "requests.cpu"    = "3000m",
        "requests.memory" = "1Gi",
        "secrets"         = "20",
        "services"        = "20"
      },
      users = [ //OBFUSCATED  ]
    }
  }
```

result after apply:

```
❯ k describe ns test-team
Name:         test-team
Labels:       kubernetes.io/metadata.name=test-team
Annotations:  <none>
Status:       Active

No resource quota.

No LimitRange resource.

---

❯ k describe ns test-team2
Name:         test-team2
Labels:       kubernetes.io/metadata.name=test-team2
              test2=test2
Annotations:  <none>
Status:       Active

Resource Quotas
  Name:            compute-quota
  Resource         Used  Hard
  --------         ---   ---
  requests.cpu     0     3
  requests.memory  0     1Gi
  Name:            object-quota
  Resource         Used  Hard
  --------         ---   ---
  secrets          2     20
  services         0     20

No LimitRange resource.
```

### DIFF
--- a/modules/aws-eks-teams/locals.tf
+++ b/modules/aws-eks-teams/locals.tf
@@ -10,4 +10,17 @@ locals {
     try(fileset(path.root, "${team_data.manifests_dir}/*"), [])
   ])
 
+  compute_hard_quota_list = ["requests.cpu", "requests.memory", "limits.cpu", "limits.memory"]
+  team_compute_hard_quotas = {
+    for team_name, team_data in var.application_teams : team_name => {
+      for quota_name in setintersection(local.compute_hard_quota_list, keys(team_data.quota)) : quota_name => team_data.quota[quota_name]
+    } if length(setintersection(local.compute_hard_quota_list, keys(try(team_data.quota, {})))) > 0
+  }
+
+  object_hard_quota_list = ["pods", "secrets", "services"]
+  team_object_hard_quotas = {
+    for team_name, team_data in var.application_teams : team_name => {
+      for quota_name in setintersection(local.object_hard_quota_list, keys(team_data.quota)) : quota_name => team_data.quota[quota_name]
+    } if length(setintersection(local.object_hard_quota_list, keys(try(team_data.quota, {})))) > 0
+  }
 }
--- a/modules/aws-eks-teams/main.tf
+++ b/modules/aws-eks-teams/main.tf
@@ -5,7 +5,7 @@ resource "kubernetes_namespace" "team" {
   for_each = var.application_teams
   metadata {
     name   = each.key
-    labels = each.value["labels"]
+    labels = lookup(each.value, "labels", {})
   }
 }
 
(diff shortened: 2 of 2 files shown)


aleixrm-scopely:
sorry about the push force noise 🙏🏽

modules/aws-eks-teams/locals.tf
```diff
@@ -10,4 +10,17 @@ locals {
     try(fileset(path.root, "${team_data.manifests_dir}/*"), [])
   ])
 
+  compute_hard_quota_list = ["requests.cpu", "requests.memory", "limits.cpu", "limits.memory"]
+  team_compute_hard_quotas = {
```
bryantbiggs, line 14: I think we can achieve this outcome without this complex logic. see below

modules/aws-eks-teams/main.tf
```diff
@@ -5,41 +5,32 @@ resource "kubernetes_namespace" "team" {
...
 # ---------------------------------------------------------------------------------------------------------------------
 resource "kubernetes_resource_quota" "team_compute_quota" {
-  for_each = var.application_teams
+  for_each = local.team_compute_hard_quotas
   metadata {
     name      = "compute-quota"
     namespace = kubernetes_namespace.team[each.key].metadata[0].name
   }
   spec {
-    hard = {
```
bryantbiggs, line 22: [hard](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/resource_quota#hard) is an optional attribute and I agree we should offer it as such. To do that, I think we can re-write this as:

```suggestion
    hard = try([each.value], [])
```

This would mean that users would need to know how to spec out the `"requests.cpu" = "4000m"`, etc., but all of that control would be left up to them. 

Thoughts @kcaws / @vara-bonthu / @Zvikan ?

Hokwang:
@bryantbiggs any updates on this issue? 
I think many things in teams module needs to update.

Zvikan:
Sorry for joining late to the party, just saw https://github.com/aws-ia/terraform-aws-eks-blueprints/issues/658 .

To keep things simple for the resource quota, what do you all think about the following approach?
```
for_each = { for team_name, team_data in var.application_teams : team_name => team_data if lookup(team_data, "quota", "") != "" }
```
Maybe there's a cleaner way to do it, but since I need that if statement to ask if `quota` was passed, I know I should expected the user to provide list of specs.

This^ allows for use-cases where customers may dont want to set resource quota at all (or maybe do it outside of the Teams module).

This^ does not fully solve the use-case where customer may want to use resource quotas but choose specific quotas (e.g. only requests.cpu , requests.memroy ,pods and services).
Edit: while writing and thinking about it, with my suggestion the workaround for customers who want to have specific set of quotas, they can create their own `kubernetes_resource_quota` resource under each team, so in order to unblock users who wants to adopt Teams but can't because of this, my suggestion above may be a good first-patch-change (and its not a breaking change).

As discussed above, either we dynamically build that spec using what this PR offers, it offers more complex logic, or, we can go the route Bryant suggested which is to try the whole block, the downside of that approach is that users will need to understand the specs and provide it properly, which we can guide via our Teams example.

Also, is there a reason we need to have two resource quotas? one for compute, one for object? both can be under a single ResourceQuota, and if so, this allows to simply pass the spec block as Bryant suggested without worrying about splitting between the two as of now.

Personally, after re-reading the official [Resource Quotas](https://kubernetes.io/docs/concepts/policy/resource-quotas/) docs, I am inclined to go with the approach that user will provide the spec, this gives the freedom and flexibility for each any use-case, and our opinionated approach to which quotas should be applied via the examples and relevant READMEs.

bryantbiggs:
closing with #678 merged, please let us know if you have any questions or feedback, thank you!

aleixrm-scopely:
Totally agree, thank you for this more clean and pragmatic solution!

Issue #576, opened 2022-05-24
[Bug]: Managed Node Group nodes fail to attach


travisnburton:
### Welcome to Amazon EKS Blueprints!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/aws-ia/terraform-aws-eks-blueprints/issues) and didn't find any.

### Amazon EKS Blueprints Release version

4.0.6

### What is your environment, configuration and the example used?

main.tf
```hcl
provider "kubernetes" {
  host                   = module.eks_blueprints.eks_cluster_endpoint
  cluster_ca_certificate = base64decode(module.eks_blueprints.eks_cluster_certificate_authority_data)

  exec {
    api_version = "client.authentication.k8s.io/v1alpha1"
    command     = "aws"
    # This requires the awscli to be installed locally where Terraform is executed
    args = ["eks", "get-token", "--cluster-name", module.eks_blueprints.eks_cluster_id, "--role-arn", var.account_access]
  }
}

provider "helm" {
  kubernetes {
    host                   = module.eks_blueprints.eks_cluster_endpoint
    cluster_ca_certificate = base64decode(module.eks_blueprints.eks_cluster_certificate_authority_data)

    exec {
      api_version = "client.authentication.k8s.io/v1alpha1"
      command     = "aws"
      # This requires the awscli to be installed locally where Terraform is executed
      args = ["eks", "get-token", "--cluster-name", module.eks_blueprints.eks_cluster_id, "--role-arn", var.account_access]
    }
  }
}


#---------------------------------------------------------------
# Example to consume eks_blueprints module
#---------------------------------------------------------------
module "eks_blueprints" {
  source = "github.com/aws-ia/terraform-aws-eks-blueprints?ref=v4.0.6"

  tags = var.tags

  # EKS Cluster VPC and Subnet mandatory config
  vpc_id             = var.vpc_id
  private_subnet_ids = var.private_subnets

  # EKS CONTROL PLANE VARIABLES
  cluster_version = "1.21"

  # AWS-AUTH ConfigMap
  map_roles = var.auth_map_roles
  map_users = var.auth_map_users

  # EKS MANAGED NODE GROUPS
  node_security_group_additional_rules = {
    # Extend node-to-node security group rules. Recommended and required for the Add-ons
    ingress_self_all = {
      description = "Node to node all ports/protocols"
      protocol    = "-1"
      from_port   = 0
      to_port     = 0
      type        = "ingress"
[...]

Issue #573, opened 2022-05-22
Error: context deadline exceeded: when destroying 'eks-cluster-with-new-vpc"


markl11:
### Welcome to Amazon EKS Blueprints!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/aws-ia/terraform-aws-eks-blueprints/issues) and didn't find any.

### Amazon EKS Blueprints Release version

https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/examples/eks-cluster-with-new-vpc

### What is your environment, configuration and the example used?

I just deployed the Terraform as instructed here:

https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/examples/eks-cluster-with-new-vpc

### What did you do and What did you see instead?

I just did 'terraform apply' which works fine, but when I try to destroy it I get the following error:

Error: context deadline exceeded

bryantbiggs:
The examples provided include all the resources to bring up the example cluster and its resources. Normally, things like VPC and cluster would be in separate statefiles. The recommendation to avoid this issue is to destroy using targets such as:
```bash
terraform destroy -target module.eks_blueprints_kuberenetes_addons
terraform destroy -target module.eks_blueprints
terraform destroy
```

kcoleman731:
@markl11 the approach is outlined in our [Getting Started guide](https://aws-ia.github.io/terraform-aws-eks-blueprints/getting-started/#cleanup). We are in the process of updating READMEs for each our our examples to reflect this approach as well.

SemperRuminare:
@kcoleman731 It looks like your example in the getting started guide has typos that say "terraform apply -target="module.eks_blueprints"" instead of terraform destroy for the last two commands.

Issue #559, opened 2022-05-17
[Bug]: Managed IAM polices are getting detached during Node groups create/update/delete


vara-bonthu:
### Welcome to Amazon EKS Blueprints!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/aws-ia/terraform-aws-eks-blueprints/issues) and didn't find any.

### Amazon EKS Blueprints Release version

v4.0.4

### What is your environment, configuration and the example used?

Managed IAM polices attached to IAM roles(IRSA) are getting detached while adding new Managed node groups, update node groups or delete node groups.  Detached Managed policies are not getting attached only when the TF apply fails during Nodegroup create, update or delete. 

e.g., Managed IAM polices -> `arn:aws:iam::aws:policy/CloudWatchAgentServerPolicy` , `arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy`

**TF Plan that shows the force replacement for IRSA Add-ons with managed policies**

```sh
  # module.eks_blueprints_kubernetes_addons.module.aws_cloudwatch_metrics[0].module.helm_addon.module.irsa[0].aws_iam_role_policy_attachment.irsa[0] must be replaced
+/- resource "aws_iam_role_policy_attachment" "irsa" {
      ~ id         = "aws001-preprod-dev-eks-cloudwatch-agent-irsa-20220517110206724700000008" -> (known after apply)
      ~ policy_arn = "arn:aws:iam::aws:policy/CloudWatchAgentServerPolicy" -> (known after apply) # forces replacement
        # (1 unchanged attribute hidden)
    }

 # module.eks_blueprints_kubernetes_addons.module.aws_vpc_cni[0].module.irsa_addon[0].aws_iam_role_policy_attachment.irsa[0] must be replaced
+/- resource "aws_iam_role_policy_attachment" "irsa" {
      ~ id         = "aws001-preprod-dev-eks-aws-node-irsa-20220517110206713100000007" -> (known after apply)
      ~ policy_arn = "arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy" -> (known after apply) # forces replacement
        # (1 unchanged attribute hidden)
    }


```

**Steps to reproduce**

**Step1:** Deploy the new [VPC example](https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/examples/eks-cluster-with-new-vpc/main.tf) 

**Step2:** Add additional Managed node group to the example and run `terraform apply` again
[...]

PR #580, opened 2022-05-26
Examples and Readme update

vara-bonthu:
### What does this PR do?

- Removed `depends_on` on `kubernetes-addons` module from all examples 
- README updated for the above examples with `--target` option
- Fixes related issues 
    - #576 
    - #559 
    - #573

### Motivation

Please read this issue for more details #559 and the motivation for this PR

### More

### For Moderators


georgelza:
not really liking the -target option, thats a band aid, that pushes this to the user, instead of the code doing this...

examples/analytics/emr-on-eks/README.md
```diff
@@ -22,45 +22,53 @@ _Note: Currently Amazon Prometheus supported only in selected regions. Please se
...
 terraform plan
 ```
 
-Deploy the pattern
+**Deploy the pattern**
+
+Deploy VPC, EKS cluster with Node groups and Kubernetes Add-ons with `--target` option
 
+```sh
+terraform apply -target="module.aws_vpc"
```
bryantbiggs, line 48: I believe for apply we want to advise a standard `terraform apply` otherwise you can get some strange results and transient failures. On the destroy side though, we definitely want to destroy addons first

vara-bonthu: Terraform apply failed in the past when we don't have `depends_on`  on the node groups. @Zvikan can add some insights here. I know it works for some examples but i am not sure if it works for all examples

Now, I have removed that `depends_on` line to avoid a major issue of replacing the managed policies. Hence i added terraform apply also with `--target`. If our E2E parallel tests confirms that it woks with one APPLY then we can change that.

I don't really like this `--target` option and we have no other option for examples as the users expect the whole example to work with one single TF APPLY. I wouldn't recommend this option for production. 

Customers/users needs to split this examples and create multiple state files for VPC, EKS & NodeGroups and Kubernetes Add-ons before they actually use in Prod.

bryantbiggs: When testing with e2e yesterday using the split up targets for applies, I would frequently see transient failures like `Error: error reading Route Table (rtb-07e46cd0c3469e326): couldn't find resource`. But using a standard `terraform apply`, the results were as expected.

However, the targeted destroy is still needed - especially for those addons that create additional AWS networking resources

Here is the current [e2e setup](https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/.github/workflows/e2e-parallel-full.yml#L63-L77) that is working as [expected](https://github.com/aws-ia/terraform-aws-eks-blueprints/actions/runs/2393294684) for reference

vara-bonthu: Thanks 👍 I will update the PR with one apply

Zvikan: Previously we were running apply with `-target` with success, I do recall seeing error like `Error: error reading Route Table (rtb-07e46cd0c3469e326): couldn't find resource` that you mentioned, but they were yet another intermittent issue.
I still think it's worth to use `--target` for apply as-well, it won't be impactful as for destroy, but it may still cause some delays (which may lead to timeout exceptions as-well) during creation.


Scenario example that I can think of is if VPC resource (e.g. NAT gateway) is taking its time to be created (for whatever reason), other modules may start create their own resources (e.g. self-managed-node) which may fail due to timeout of creation which is mainly caused by NAT, as the nodes were not able to connect to the cluster for example.

Will take a look at the failures around the "route table can't find resource issue" if it's reproducible more easily today.

examples/analytics/emr-on-eks/README.md
```diff
@@ -128,6 +128,22 @@ cd examples/analytics/emr-on-eks/examples/
 ./delete_emr_virtual_cluster_for_eks.sh
 ```
 
+To clean up your environment, destroy the Terraform modules in reverse order.
```
bryantbiggs, line 131: I can't comment since the lines not changed but line 122 should change to `## Step 6: Cleanup`

examples/analytics/emr-on-eks/README.md
```diff
@@ -128,6 +128,22 @@ cd examples/analytics/emr-on-eks/examples/
...
+
+Destroy the Kubernetes Add-ons, EKS cluster with Node groups and VPC
+
+```sh
+terraform destroy -target="module.eks_blueprints_kubernetes_addons"
```
bryantbiggs, line 136: I think we want to add `-auto-approve` to all of these

PR #597, opened 2022-06-05
fix: typo in null_resource

niv1612:
### What does this PR do?

fixing a typo: from kube_system to kube-system

### Motivation

Right now, fargate-serverless doesn't work well. This type prevents it from running.

### DIFF
--- a/examples/fargate-serverless/main.tf
+++ b/examples/fargate-serverless/main.tf
@@ -194,7 +194,7 @@ resource "null_resource" "modify_kube_dns" {
 
     # We are maintaing the existing kube-dns service and annotating it for Helm to assume control
     command = <<-EOT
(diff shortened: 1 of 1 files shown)


examples/fargate-serverless/main.tf
```diff
@@ -194,7 +194,7 @@ resource "null_resource" "modify_kube_dns" {
 
     # We are maintaing the existing kube-dns service and annotating it for Helm to assume control
     command = <<-EOT
-      echo "Setting implicit dependency on ${module.eks_blueprints.fargate_profiles["kube_system"].eks_fargate_profile_arn}"
```
bryantbiggs, line 197: Thanks for the PR! However, this is correct as written - its referring to the Fargate profiles map input key which is `kube_system` https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/f5d31cf7a39e343f2c571e264dd0b7c13299abd3/examples/fargate-serverless/main.tf#L75

PR #606, opened 2022-06-07
Using EKS Add-on to configure AWS Distro for OpenTelemetry

saaish:
### More

### For Moderators


modules/kubernetes-addons/opentelemetry-operator/main.tf
```diff
@@ -1,10 +1,168 @@
...
+resource "aws_eks_addon" "example" {
+  cluster_name = var.addon_context.eks_cluster_id
+  resolve_conflicts = "OVERWRITE"
+  addon_name   = "adot"
+  depends_on = [kubectl_manifest.opentelemetry-prereq]
+}
 
-  depends_on = [kubernetes_namespace_v1.prometheus]
+resource "kubectl_manifest" "opentelemetry-prereq" {
+    yaml_body = <<YAML
```
askulkarni2, line 9: Aesthetically speaking, I think its best if separate this out into yaml files and read them in via `templatefile()`.

askulkarni2 (changes requested):
@saaish thanks for the PR. This gets rid of the self-managed add-on. Is this what we are going to recommend our customers? Should they have an option to use either EKS managed or self-managed?

A couple of things to address:
1. Add description, motivation, details on tests performed.
2. Fix the broken CI issues.

modules/kubernetes-addons/opentelemetry-operator/main.tf
bryantbiggs: This is on me - I asked @saaish to push up what he had so far so I could take a look but I got swamped with other requests. I'll take a look and update

bryantbiggs:
@saaish looking into this further, I think we might be adopting this a bit too soon. It looks like right now there is a compatibility issue with cert-manager v1.6+ and currently we default to v1.7.1 - In addition, it looks like we have to create the RBAC permissions which isn't a great "managed" experience. what improvements can users expect with the EKS addon for ADOT over the self-managed addon via Helm?

PR #617, opened 2022-06-09
Remove hardcoded regions

bonclay7:
### What does this PR do?

AWS regions settings were hardcoded in some samples, confusing some users

### More

- [x] Yes, I have tested the PR using my local account setup  (Provide any test evidence report under Additional Notes)
- [x] Yes, I ran `pre-commit run -a` with this PR

**Note**: Not all the PRs required examples and docs except a new pattern or add-on added.

### For Moderators
- [ ] E2E Test successfully complete before merge?

PR #627, opened 2022-06-12
Another Proposal for #606

bonclay7:
### What does this PR do?
Managed Add-on for ADOT (started in #606)

### More

### For Moderators


examples/observability/adot-amp-grafana-for-haproxy/main.tf
```diff
@@ -82,7 +94,6 @@ module "eks_blueprints_kubernetes_addons" {
   eks_oidc_provider    = module.eks_blueprints.oidc_provider
   eks_cluster_version  = module.eks_blueprints.eks_cluster_version
 
-  enable_cert_manager                  = true
```
Zvikan, line 85: what's the reason to remove this?

examples/observability/adot-amp-grafana-for-java/main.tf
```diff
@@ -37,7 +49,7 @@ data "aws_availability_zones" "available" {}
 
 locals {
   name   = basename(path.cwd)
-  region = "us-west-2"
+  region = var.aws_region
```
Zvikan, line 52: `aws_region` variable is not defined on this PR, is that suppose to be part of https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/617 or something else?

modules/kubernetes-addons/opentelemetry-operator/main.tf
```diff
@@ -1,17 +1,33 @@
-module "operator" {
-  source        = "../helm-addon"
-  helm_config   = local.helm_config
-  irsa_config   = null
+module "cert_manager" {
```
Zvikan, line 1: this may be related to my question above about removing cert_manager, can you share more info about this change? what we did before and what we propose to do now.

Zvikan (commented):
Thank you for the PR, can you please fill the PR template and share more insights about the changes and mainly the motivation behind the changes, I see reference to PR #606  which is not detailed, and some changes that may be related to a different PR (https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/617).

bryantbiggs:
@bonclay7 could you answer this please https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/606#issuecomment-1152939955

modules/kubernetes-addons/opentelemetry-operator/main.tf
bonclay7: Fair point -- I included cert manager inside the opentelemetry-operator module because it is an absolute pre-condition for the add-on. That being said, I have tried a dependency on the kubernetes operator main file, however, I encountered a terraform blocker where `for_each` (which is used to split permissions manifests) and `count` are used. 
If you have another suggetion, I'll take it

examples/observability/adot-amp-grafana-for-java/main.tf
bonclay7: Yes indeed, this is part of #617, I'll remove here

bonclay7:
Thanks for the review, please check changes/comments @Zvikan @bryantbiggs

modules/kubernetes-addons/main.tf
```diff
@@ -332,6 +332,8 @@ module "adot_collector_java" {
   amazon_prometheus_workspace_endpoint = var.amazon_prometheus_workspace_endpoint
   amazon_prometheus_workspace_region   = var.amazon_prometheus_workspace_region
   addon_context                        = local.addon_context
+
+  depends_on = [module.opentelemetry_operator]
```
bryantbiggs, line 336: what happens when the operator is not installed/available first? will this eventually reconcile and therefore we can remove the `depends_on`? Same for the other collectors below

modules/kubernetes-addons/opentelemetry-operator/main.tf
```diff
@@ -1,17 +1,33 @@
...
+  source        = "../cert-manager"
   addon_context = var.addon_context
-
-  depends_on = [kubernetes_namespace_v1.prometheus]
 }
 
-resource "kubernetes_namespace_v1" "prometheus" {
+resource "kubernetes_namespace_v1" "adot" {
   metadata {
     name = local.helm_config["namespace"]
```
bryantbiggs, line 8: we have to be careful with this - if the EKS addon for ADOT expects a [specific namespace](https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/627/files#diff-0393764fcbfc5379a2df275af001d20d3e8c190ab4570efcb83340c3eeee8bc1R5) that is not configurable, then we have to match that. Do checkout the CoreDNS addon where we have two distinct set of values - one for EKS addon, one for self-managed addon. This ensures that if the users selects the self-managed route, we can honor that flexibility, but if they select the EKS addon, we have to match what the EKS addon expects

modules/kubernetes-addons/opentelemetry-operator/main.tf
```diff
@@ -1,17 +1,33 @@
...
+data "kubectl_path_documents" "adot" {
+  pattern = "${path.module}/manifests/*.yaml"
+}
+
+# official kubernetes_manifest issue
+# https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/manifest#before-you-use-this-resource
+resource "kubectl_manifest" "adot" {
+  for_each   = toset(data.kubectl_path_documents.adot.documents)
+  yaml_body  = each.value
+  depends_on = [kubernetes_namespace_v1.adot]
```
bryantbiggs, line 25: This is just an FYI - we generally want to avoid `depends_on` as much as possible. To get around this, look for the "breadcrumbs" that you can feed into a child module/resource to make it implicitly dependent on the parent module/resource. Deployment order is a core tenant of Terraform and providing that information to Terraform will ensure things are deployed at the optimal time without unintended side-effects (explicit dependencies will cause the child resource/module to re-compute all computed attributes when the parent resource/module is updated - this can lead to some really bad experiences and infra service disruptions)


1. Yes, as you've discovered the `kubernetes_manifest` resource has some gnarly limitations. However, we don't need to use that here - we can use the specific resource types like `kubernetes_cluster_role_v1`,  `kubernetes_cluster_role_binding_v1`, etc., or we can just merge the manifests into one file like the AWS docs recommend and provide the [link](https://amazon-eks.s3.amazonaws.com/docs/addons-otel-permissions.yaml) for and just do a single `kubectl_manifest` resource without the `for_each` loop. 
2. Going back to the point of "breadcrumbs", if we create a template variable in the manifest file for [`namespace`](https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/627/files#diff-0393764fcbfc5379a2df275af001d20d3e8c190ab4570efcb83340c3eeee8bc1R5), we could potentially get two wins - ensure the namespaces are synced across resources, ensure the proper deployment order (terraform will see this input from the `kubernetes_namespace_v1` resource and know it needs to be created first and we can drop the `depends_on`)

modules/kubernetes-addons/opentelemetry-operator/main.tf
```diff
@@ -1,17 +1,33 @@
...
+}
+
+resource "aws_eks_addon" "adot" {
```
bryantbiggs, line 28: Refer to the current structure for [CoreDNS](https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/modules/kubernetes-addons/aws-coredns/main.tf#L5-L26) so we match that same level of functionality.

bryantbiggs:
@bonclay7 FYI I am working on the review and proposed changes - just taking longer than expected. I hope to have something for your review later today

modules/kubernetes-addons/main.tf
```diff
@@ -333,6 +333,15 @@ module "opentelemetry_operator" {
   source        = "./opentelemetry-operator"
   helm_config   = var.opentelemetry_operator_helm_config
   addon_context = local.addon_context
+
+  # Amazon EKS ADOT add
```
bryantbiggs, line 337: Don't forget we'll need to support both the EKS ADOT route as well as the current self-managed otel operator. See [here](https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/639/files#diff-ae264c430a0512ec2b96032bc8a245bb09a2da086fc5edd677c1d31fb25dca71R324-R339) for reference

bonclay7: Yes! just updated this

modules/kubernetes-addons/opentelemetry-operator/main.tf
```diff
@@ -1,14 +1,288 @@
...
+
+module "helm_addon" {
+  source = "../helm-addon"
+  count  = var.enable_opentelemetry_operator ? 1 : 0
+
+  helm_config   = local.helm_config
+  irsa_config   = null
+  addon_context = var.addon_context
+
+  depends_on = [module.cert_manager]
```
bonclay7, line 287: @bryantbiggs Tried the cluster id output from cert manager, but it doesn't work consistently. For example when the value is already in the tfstate, TF doesn't prioritize cert-manager anymore

modules/kubernetes-addons/main.tf
```diff
@@ -329,44 +329,92 @@ module "velero" {
...
-  source                               = "./adot-collector-java"
-  helm_config                          = var.adot_collector_java_helm_config
+  count  = var.enable_adot_collector_java ? 1 : 0
+  source = "./adot-collector-java"
+
+  helm_config = merge(
+    var.adot_collector_java_helm_config,
+    {
+      # Using the output from the operator addon, this will ensure correct deployment order
+      operator_namespace = module.opentelemetry_operator[0].namespace
```
bonclay7, line 358: @bryantbiggs this doesn't produce a consistent behaviour, as it tries sometimes to install the app before the collector

bryantbiggs: oy, ok - I guess we can abandon this and use the hard dependency with `depends_on`

bryantbiggs (approved):

PR #634, opened 2022-06-14
Update addons versions | Examples to use cluster version 1.22 | Fix self-managed-node example

Zvikan:
### What does this PR do?

- Update all addons to latest released version 
- Update all examples cluster_version to 1.22
- Add helm provider to self-manged node example

### Motivation

- This industry moves fast. Addons contributors fixing bugs, adding features, resolving security concerns and more! we need to keep things updated.
- Support EKS version 1.22 for EKS Blueprints
  - NOTE: this PR does not cover migration from 1.21 to 1.22, this is focused on testing and validating creation of clusters versions 1.22 and deploying addons using the latest version across all current examples ( see below "For Moderators" for link).
- self-managed node group is failing today due to helm provider config missing, preventing from issues such as:
```
Error: Kubernetes cluster unreachable: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable

```
### More

   - Sanity testing by leveraging E2E action workflow, creating examples and leaving clusters up, then manually checking each individual cluster events, nodes and overall pods restarts.

### For Moderators
- [X] E2E Test successfully complete before merge? - https://github.com/aws-ia/terraform-aws-eks-blueprints/actions/runs/2513475993 

### Additional Notes

- This PR focused around cluster creation using version 1.22 with latest addons versions, this does not cover upgrade from 1.21 to 1.22.

### DIFF
--- a/examples/analytics/emr-on-eks/main.tf
+++ b/examples/analytics/emr-on-eks/main.tf
@@ -50,7 +50,7 @@ module "eks_blueprints" {
   source = "../../.."
 
   cluster_name    = local.name
-  cluster_version = "1.21"
+  cluster_version = "1.22"
 
   vpc_id             = module.vpc.vpc_id
   private_subnet_ids = module.vpc.private_subnets
--- a/examples/analytics/spark-k8s-operator/main.tf
+++ b/examples/analytics/spark-k8s-operator/main.tf
@@ -50,7 +50,7 @@ module "eks_blueprints" {
   source = "../../.."
 
   cluster_name    = local.name
-  cluster_version = "1.21"
+  cluster_version = "1.22"
 
   vpc_id             = module.vpc.vpc_id
   private_subnet_ids = module.vpc.private_subnets
--- a/examples/aws-efs-csi-driver/main.tf
+++ b/examples/aws-efs-csi-driver/main.tf
@@ -50,7 +50,7 @@ module "eks_blueprints" {
   source = "../.."
 
   cluster_name    = local.name
-  cluster_version = "1.21"
+  cluster_version = "1.22"
 
   vpc_id             = module.vpc.vpc_id
   private_subnet_ids = module.vpc.private_subnets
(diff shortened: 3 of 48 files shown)


Zvikan:
Additional changes pushed to the branch:
- Set cluster version to 1.22
- Pushed fix for self-managed node example, helm provider is missing and without it example is failing
```
Error: Kubernetes cluster unreachable: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
```


What I noticed so far by testings our e2e parallel with the above changes:
1. We are currently using single terraform apply & destroy command , this leads to timeout failures due to addons hitting timeout, this mainly caused because worker nodes and addons are created in parallel by terraform, but it can take around 10-15 minutes for worker nodes to be created and join the cluster, and until then addons will be hanging with "Creating" state.
Ideally, addons should be start be in status "Creating" only after worker nodes (AKA blueprints module) completed successfully, bringing back `-target` to the E2E may be the right idea.
2. Self-managed addon failing as called out above, this is probably true also to current `main` branch and this PR also fixing this for now.
3. Karpenter example is failing due to destroy issue with NTH eventbridge rules, where rules cant be created because there are associated targets...diving deep into this issue I figured out that the targets are not related to the karpenter example but other example, which lead me to see that we're currently creating the [same event rules names](https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/modules/kubernetes-addons/aws-node-termination-handler/locals.tf#L44-L76), so this means if I have 2 clusters that have enabled NTH, the event bridge rules will get mixed and TF won't know about it (there's no failures like "This event rule name already exists" it seems).
A suggestion would be to give the event rule a unique name, e.g. keeping same names with postfix of the cluster name.
Changes has not been done to this yet.

link to E2E workflows: https://github.com/aws-ia/terraform-aws-eks-blueprints/actions/runs/2512350932 (note failures re-ran).


So far I have no identified any "sanity" failures (i.e. terraform apply and destroy works) related to 1.22 or addons with latest version.

bryantbiggs (approved):
awesome, thank you!

kcoleman731 (commented):
Looks great thank you!

jybaek:
@Zvikan This code has inspired me a lot, thanks. If I make it work with GitHub Actions, I need to install the AWS CLI first, right?

PR #657, opened 2022-06-20
fix: Allow settting metadata endpoint options so users can configure or disable

bryantbiggs:
### What does this PR do?
- Allow setting metadata endpoint options so users can configure or disable

### Motivation
- Allow users to configure the metadata endpoint options to suit their workloads, or even disable if not supported in their region

### More

### For Moderators


kcoleman731 (commented):
Looks great. Can you add customization to one of our examples?

modules/aws-eks-managed-node-groups/managed-launch-templates.tf
```diff
@@ -34,9 +34,9 @@ resource "aws_launch_template" "managed_node_groups" {
   }
 
   metadata_options {
-    http_endpoint               = try(var.context.http_endpoint, "enabled")
-    http_tokens                 = try(var.context.http_tokens, "required") #tfsec:ignore:aws-autoscaling-enforce-http-token-imds
-    http_put_response_hop_limit = try(var.context.http_put_response_hop_limit, 2)
+    http_endpoint               = try(var.managed_ng.http_endpoint, "enabled")
+    http_tokens                 = try(var.managed_ng.http_tokens, "required") #tfsec:ignore:aws-autoscaling-enforce-http-token-imds
+    http_put_response_hop_limit = try(var.managed_ng.http_put_response_hop_limit, 2)
```
vara-bonthu, line 39: This will still work but for consistency, let' use the merged config local variable

```hcl
http_endpoint               = try(local.managed_node_group["http_endpoint"], "enabled")
http_tokens                 = try(local.managed_node_group["http_tokens"], "required") #tfsec:ignore:aws-autoscaling-enforce-http-token-imds
http_put_response_hop_limit = try(local.managed_node_group["http_put_response_hop_limit"], 2)
```

modules/aws-eks-self-managed-node-groups/self-managed-launch-templates.tf
```diff
@@ -17,9 +17,9 @@ module "launch_template_self_managed_ng" {
       kubelet_extra_args   = local.self_managed_node_group["kubelet_extra_args"]
       monitoring           = local.self_managed_node_group["enable_monitoring"]
 
-      http_endpoint               = var.context.http_endpoint
-      http_tokens                 = var.context.http_tokens
-      http_put_response_hop_limit = var.context.http_put_response_hop_limit
+      http_endpoint               = try(var.self_managed_ng.http_endpoint, "enabled")
+      http_tokens                 = try(var.self_managed_ng.http_tokens, "required")
+      http_put_response_hop_limit = try(var.self_managed_ng.http_put_response_hop_limit, 2)
```
vara-bonthu, line 22: Same as above comment

modules/aws-eks-managed-node-groups/managed-launch-templates.tf
bryantbiggs: Isn't this redundant? I just removed the redundancy while keeping the default settings

vara-bonthu: I meant either we have to use `try` for all the defaults or use the existing approach with **merged map** variable(`local.managed_node_group`) but not both.

examples/node-groups/managed-node-groups/main.tf
```diff
@@ -134,6 +134,10 @@ module "eks_blueprints" {
       eni_delete        = true
       public_ip         = false # Use this to enable public IP for EC2 instances; only for public subnets used in launch templates
 
+      http_endpoint               = "enabled"
+      http_tokens                 = "optional"
+      http_put_response_hop_limit = 3
```
askulkarni2, line 139: Should we change this in another example or create a new one? We probably want the default MNG example to use the default options.

modules/launch-templates/main.tf
```diff
@@ -76,10 +76,14 @@ resource "aws_launch_template" "this" {
...
-    http_endpoint               = each.value.http_endpoint
-    http_tokens                 = each.value.http_tokens
-    http_put_response_hop_limit = each.value.http_put_response_hop_limit
+  dynamic "metadata_options" {
+    for_each = try(each.value.enable_metadata_options, true) ? [1] : []
+
+    content {
+      http_endpoint               = try(each.value.http_endpoint, "enabled")
+      http_tokens                 = try(each.value.http_tokens, "required")
+      http_put_response_hop_limit = try(each.value.http_put_response_hop_limit, 2)
```
askulkarni2, line 85: Should we add [ipv6 support](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#http_protocol_ipv6) as well?

examples/node-groups/managed-node-groups/main.tf
bryantbiggs: I believe [this](https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/657/files/698fea88e7db215cb116aeddfa676ad51edea4d4#diff-a74a95b7fe2f941d719e545e427114d9c5b80d1131125d8d2d698b35249004ebR107) is the "default" or minimum config node group. I added to this node group since it looked like it was demonstrating all the options available, but I am happy to move to a different node group example

vara-bonthu (approved):

PR #691, opened 2022-06-26
[teams module] add iam policy for application team

Hokwang:
### What does this PR do?

- add iam policy for application team
- adjust resource names

### Motivation

- issue 647 Q3
- some resource name is just team,

### More

- [x] Yes, I have created another PR for add-ons under [add-ons](https://github.com/aws-samples/eks-blueprints-add-ons) repo (if applicable)

### For Moderators

### DIFF
--- a/modules/aws-eks-teams/main.tf
+++ b/modules/aws-eks-teams/main.tf
@@ -228,7 +228,7 @@ resource "aws_iam_policy" "platform_team" {
   name        = "${var.eks_cluster_id}-PlatformTeamEKSAccess"
   path        = "/"
   description = "Platform Team EKS Access IAM Policy"
-  policy      = var.platform_team_iam_policy == "" ? data.aws_iam_policy_document.platform_team_default.json : var.platform_team_iam_policy
+  policy      = var.platform_team_iam_policy == "{}" ? data.aws_iam_policy_document.platform_team_default.json : var.platform_team_iam_policy
 
   tags = var.tags
 }
--- a/modules/aws-eks-teams/main.tf
+++ b/modules/aws-eks-teams/main.tf
@@ -172,8 +172,8 @@ resource "kubernetes_service_account" "application_team_sa" {
 }
 
 # Kubernetes Manifests
-resource "kubectl_manifest" "team" {
-  for_each = { for manifest in local.team_manifests : manifest => file(manifest) }
+resource "kubectl_manifest" "application_team_manifest" {
+  for_each = { for manifest in local.application_team_manifests : manifest => file(manifest) }
   
   yaml_body = each.value
   
--- a/modules/aws-eks-teams/main.tf
+++ b/modules/aws-eks-teams/main.tf
@@ -228,7 +228,7 @@ resource "aws_iam_policy" "platform_team" {
   name        = "${var.eks_cluster_id}-PlatformTeamEKSAccess"
   path        = "/"
   description = "Platform Team EKS Access IAM Policy"
-  policy      = var.platform_team_iam_policy == "" ? data.aws_iam_policy_document.platform_team_iam_policy[0].json : var.platform_team_iam_policy
+  policy      = var.platform_team_iam_policy == "" ? data.aws_iam_policy_document.platform_team_default[0].json : var.platform_team_iam_policy
 
   tags = var.tags
 }
--- a/modules/aws-eks-teams/main.tf
+++ b/modules/aws-eks-teams/main.tf
@@ -1,23 +1,20 @@
-# ---------------------------------------------------------------------------------------------------------------------
 # Namespace
-# ---------------------------------------------------------------------------------------------------------------------
-resource "kubernetes_namespace" "team" {
+resource "kubernetes_namespace" "application_team_ns" {
   for_each = var.application_teams
+  
   metadata {
     name   = each.key
     labels = try(each.value["labels"], {})
   }
 }
 
-# ---------------------------------------------------------------------------------------------------------------------
 # Quotas
-# ---------------------------------------------------------------------------------------------------------------------
-resource "kubernetes_resource_quota" "this" {
+resource "kubernetes_resource_quota" "application_team_quota" {
   for_each = var.application_teams
 
   metadata {
-    name      = "quotas"
-    namespace = kubernetes_namespace.team[each.key].metadata[0].name
+    name      = "quota"
+    namespace = kubernetes_namespace.application_team_ns[each.key].metadata[0].name
   }
 
   spec {
@@ -25,14 +22,13 @@ resource "kubernetes_resource_quota" "this" {
   }
 }
 
-# ---------------------------------------------------------------------------------------------------------------------
-# IAM / RBAC
-# ---------------------------------------------------------------------------------------------------------------------
-resource "aws_iam_role" "team_access" {
+# Application Team EKS Access IAM Role
+resource "aws_iam_role" "application_team_iam_role" {
   for_each = { for team_name, team_data in var.application_teams : team_name => team_data if lookup(team_data, "users", "") != "" }
 
-  name                 = "${var.eks_cluster_id}-${each.key}-access"
+  name                 = "${var.eks_cluster_id}-${each.key}-application-team-iam-role"
   permissions_boundary = var.iam_role_permissions_boundary
+  managed_policy_arns  = [aws_iam_policy.application_team_iam_policy[0].arn]
 
   assume_role_policy = jsonencode({
     "Version" : "2012-10-17",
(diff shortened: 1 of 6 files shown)


modules/aws-eks-teams/main.tf
```diff
@@ -98,29 +111,30 @@ resource "kubernetes_role" "team" {
   }
 }
 
-resource "kubernetes_role_binding" "team" {
+resource "kubernetes_role_binding" "application_team_role_binding" {
```
Zvikan, line 114: same as above

modules/aws-eks-teams/main.tf
```diff
@@ -1,38 +1,34 @@
-# ---------------------------------------------------------------------------------------------------------------------
```
Zvikan, line 1: Is there a reason why we're deleting all the nice and structured comments :) ?

modules/aws-eks-teams/main.tf
```diff
@@ -145,42 +159,36 @@ resource "aws_iam_role" "team_sa_irsa" {
   tags = var.tags
 }
 
-# ---------------------------------------------------------------------------------------------------------------------
-# Kubernetes Team Service Account
```
Zvikan, line 149: Same as above, the reason for having this bulky structured comments it to allow users to read between the lines and see as sections ("This is for IRSA", "This is For Kubernetes Service Accounts").
The concept of Teams may be new to some, and many would like to dive deep into the code and understand its structure, having this can help navigate through the code more easily (remember, not everyone are terraform experts).

modules/aws-eks-teams/main.tf
```diff
@@ -198,15 +206,14 @@ resource "aws_iam_role" "platform_team" {
...
-# Platform Team EKS access IAM policy
-# ---------------------------------------------------------------------------------------------------------------------
+# Platform Team EKS Access IAM policy
+resource "aws_iam_policy" "platform_team_iam_policy" {
+  count = length(var.platform_teams) > 0 ? 1 : 0
 
-resource "aws_iam_policy" "platform_team_eks_access" {
-  count       = length(var.platform_teams) > 0 ? 1 : 0
-  name        = "${var.eks_cluster_id}-PlatformTeamEKSAccess"
+  name        = "${var.eks_cluster_id}-platform-team-iam-policy"
```
Zvikan, line 213: I think the previous name was good and followed AWS policy naming convention, it was clear defining the purpose of the policy.
What is the reason for the suggested change?

modules/aws-eks-teams/outputs.tf
```diff
@@ -1,41 +1,41 @@
...
   })
 }
 
-output "application_teams_configure_kubectl" {
-  description = "Configure kubectl for each Application Teams: make sure you're logged in with the correct AWS profile and run the following command to update your kubeconfig"
+output "platform_team_configure_kubectl" {
+  description = "Configure kubectl for each Platform Team: make sure you're logged in with the correct AWS profile and run the following command to update your kubeconfig"
   value = tomap({
-    for k, v in aws_iam_role.team_access : k => "aws eks --region ${data.aws_region.current.id} update-kubeconfig --name ${data.aws_eks_cluster.eks_cluster.name}  --role-arn ${v.arn}"
+    for k, v in aws_iam_role.platform_team_iam_role : k => "aws eks --region ${data.aws_region.current.id} update-kubeconfig --name ${data.aws_eks_cluster.eks_cluster.name}  --role-arn ${v.arn}"
```
Zvikan, line 39: note: changing output names is considered breaking change.
If we believe the original output names are not clear enough or confusing, we should change it with clear callout on next release for customer to know that we did it and what they should change from their end.

modules/aws-eks-teams/main.tf
```diff
@@ -1,38 +1,34 @@
...
-# ---------------------------------------------------------------------------------------------------------------------
-resource "aws_iam_role" "team_access" {
+# Application Team EKS Access IAM Role
+resource "aws_iam_role" "application_team_iam_role" {
   for_each = { for team_name, team_data in var.application_teams : team_name => team_data if lookup(team_data, "users", "") != "" }
 
-  name                 = "${var.eks_cluster_id}-${each.key}-access"
+  name                 = "${var.eks_cluster_id}-${each.key}-application-team-iam-role"
   permissions_boundary = var.iam_role_permissions_boundary
+  managed_policy_arns  = [aws_iam_policy.application_team_iam_policy[0].arn]
```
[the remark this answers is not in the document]
Hokwang: I think user can declare many permissions in one policy, there's no problem create one policy.

modules/aws-eks-teams/outputs.tf
Hokwang: I agreed so roll back to previous version. But you need to make breaking change for this output in near future...
I think we need to big changes for teams module for my Question 1.

Hokwang:
@Zvikan Please re-review when you are available.

Zvikan:
> @Zvikan Please re-review when you are available.

Thank you, can you please check the failed CI checks, related to the teams example:
```
Error: "policy" contains an invalid JSON policy
  with module.eks_blueprints.module.aws_eks_teams[0].aws_iam_policy.application_team[0],
  on ../../modules/aws-eks-teams/main.tf line 64, in resource "aws_iam_policy" "application_team":
  64:   policy      = var.application_team_iam_policy
```
https://github.com/aws-ia/terraform-aws-eks-blueprints/runs/7155391530?check_suite_focus=true

Hokwang:
@Zvikan sorry for late, updated.

modules/aws-eks-teams/main.tf
```diff
@@ -98,29 +119,30 @@ resource "kubernetes_role" "team" {
...
     api_group = "rbac.authorization.k8s.io"
     kind      = "Role"
     name      = "${each.key}-role"
   }
   subject {
+    api_group = "rbac.authorization.k8s.io"
     kind      = "Group"
     name      = "${each.key}-group"
-    api_group = "rbac.authorization.k8s.io"
-    namespace = kubernetes_namespace.team[each.key].metadata[0].name
```
Zvikan, line 116: Namespace not required for group subject anymore? it was never needed?

modules/aws-eks-teams/main.tf
```diff
@@ -146,41 +168,40 @@ resource "aws_iam_role" "team_sa_irsa" {
...
   for_each = var.application_teams
+
   metadata {
     name        = format("%s-sa", each.key)
-    namespace   = kubernetes_namespace.team[each.key].metadata[0].name
-    annotations = { "eks.amazonaws.com/role-arn" : aws_iam_role.team_sa_irsa[each.key].arn }
+    namespace   = kubernetes_namespace.this[each.key].metadata[0].name
+    annotations = { "eks.amazonaws.com/role-arn" : aws_iam_role.application_team_irsa[each.key].arn }
   }
-  automount_service_account_token = true
```
Zvikan, line 159: Removed because defaults to true?

modules/aws-eks-teams/main.tf
```diff
@@ -50,10 +52,25 @@ resource "aws_iam_role" "team_access" {
...
+  count = length(var.application_teams) > 0 ? 1 : 0
+
+  name        = "${var.eks_cluster_id}-ApplicationTeamEKSAccess"
+  path        = "/"
+  description = "Application Team EKS Access IAM Policy"
+  policy      = var.application_team_iam_policy
+
+  tags = var.tags
+}
+
```
Zvikan, line 68: Can you please share more about this use case?
I assume you wanted to provide EKS Access (UI) to your application team, but do you also want to provide custom policy for your use case?
And if so, this means that every team using the same module will have this policy applied to the role. the name of the role and the policy provided is around "Access", but access to what?
So, the way I think about this is:
1. we can have similar "access" to EKS UI as the platform team (need to confirm that the team will only see resources relevant to the team via the UI, should be since it's provided via the k8s auth map setup) with default policy or allow users to provide custom one.
Also note the name of the policy which is "ApplicationTeamEKSAccess".
2. on line 37 here, we apply the same provided policy to all the teams, this might not be what customer expect to happen.

modules/aws-eks-teams/main.tf
Hokwang: yes

modules/aws-eks-teams/main.tf
Hokwang: true!

modules/aws-eks-teams/main.tf
Hokwang: 1. Actually this is not related with console UI access for team(person), but I followed iam_policy name for platform team.

I need to grant some permissions to application, so it's serviceaccount, for example,
access s3, update cloudfront distribution, create route53 hostedzone, etc.
This web application uses/controls AWS resources.

2. In my original issue, this is related with Q1. In my case, one application team use two namespaces,
so this is okay for now.

I think that your team need to decide whether to allow multiple application teams to exist,
and they can use one or multiple namespaces, and how to implement this. That will be a fundamental solution.

Zvikan: Maybe I misunderstood something but If your use case is for web application, why this is related to Teams?
Teams by design is for human access and interaction with the cluster, allows a team to access their "environment" (AKA namespace/s).

I disagree with point number 2,  we should think big here for overall usage, going with this direction limits any other customers who want to use this module where a single policy will always apply to all other application teams, this is very opinionated approach that we want to be careful with. 
We should have a better design to approach your ask on point 2 IMO, this current approach could be a feature flag to apply same policy for all application teams.

bryantbiggs:
hi @Hokwang - thank you for the PR. We are currently reviewing the feedback and usage of the `teams` module and will be looking at making some changes to it. For now we are going to close this PR and we encourage you to share any feedback on #842 while we deliberate on how we evolve the module to better suit users

PR #698, opened 2022-06-27
feat: add upstream module with k8s-addons example

Zvikan:
### What does this PR do?

- Adds new example using eks upstream module("terraform-aws-modules/eks/aws") with blueprints addons module only

### Motivation

- https://github.com/aws-ia/terraform-aws-eks-blueprints/issues/689 

### More

### For Moderators

### DIFF
--- a/examples/upstream-with-k8s-addons/main.tf
+++ b/examples/upstream-with-k8s-addons/main.tf
@@ -73,7 +73,7 @@ module "eks" {
       desired_size = 1
 
       update_config = {
-        max_unavailable_percentage = 75
+        max_unavailable_percentage = 33
       }
     }
   }
--- a/examples/upstream-with-k8s-addons/outputs.tf
+++ b/examples/upstream-with-k8s-addons/outputs.tf
@@ -1,34 +1,4 @@
-output "vpc_private_subnet_cidr" {
-  description = "VPC private subnet CIDR"
-  value       = module.vpc.private_subnets_cidr_blocks
-}
-
-output "vpc_public_subnet_cidr" {
-  description = "VPC public subnet CIDR"
-  value       = module.vpc.public_subnets_cidr_blocks
-}
-
-output "vpc_cidr" {
-  description = "VPC CIDR"
-  value       = module.vpc.vpc_cidr_block
-}
-
-output "eks_cluster_id" {
-  description = "EKS cluster ID"
-  value       = module.eks.cluster_id
-}
-
-output "eks_managed_node_groups" {
-  description = "EKS managed node groups"
-  value       = module.eks.eks_managed_node_groups
-}
-
 output "configure_kubectl" {
   description = "Configure kubectl: make sure you're logged in with the correct AWS profile and run the following command to update your kubeconfig"
   value       = "aws eks --region ${local.region} update-kubeconfig --name ${module.eks.cluster_id}"
-}
-
-output "region" {
-  value       = local.region
-  description = "AWS region"
 }
--- a/examples/upstream-with-k8s-addons/versions.tf
+++ b/examples/upstream-with-k8s-addons/versions.tf
@@ -20,6 +20,6 @@ terraform {
   # backend "s3" {
   #   bucket = "terraform-ssp-github-actions-state"
   #   region = "us-west-2"
-  #   key    = "e2e/eks-cluster-with-new-vpc/terraform.tfstate"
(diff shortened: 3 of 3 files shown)


examples/upstream-with-k8s-addons/main.tf
```diff
@@ -0,0 +1,144 @@
...
+# Kubernetes Addons using Blueprints kubernetes-addons module
+#---------------------------------------------------------------
+
+module "eks_blueprints_kubernetes_addons" {
+  source = "../../modules/kubernetes-addons"
+
+  eks_cluster_id       = module.eks.cluster_id
+  eks_cluster_endpoint = module.eks.cluster_endpoint
+  eks_oidc_provider    = module.eks.oidc_provider
+  eks_cluster_version  = module.eks.cluster_version
```
Zvikan, line 92: This is really the most important part of this example, should we call this out in the README/add a comment here?

examples/upstream-with-k8s-addons/main.tf
```diff
@@ -0,0 +1,144 @@
...
+    Blueprint  = local.name
+    GithubRepo = "github.com/aws-ia/terraform-aws-eks-blueprints"
+  }
+}
+
+#---------------------------------------------------------------
+# EKS Cluster with terraform-aws-eks module
+#---------------------------------------------------------------
+
+module "eks" {
```
Zvikan, line 50: Tried to keep the example as simple as possible, if you believe we can further remove more pieces here lmk.

examples/upstream-with-k8s-addons/main.tf
```diff
@@ -0,0 +1,144 @@
...
+    bottlerocket = {
+      ami_type = "BOTTLEROCKET_x86_64"
+      platform = "bottlerocket"
+
+      min_size     = 1
+      max_size     = 7
+      desired_size = 1
+
+      update_config = {
+        max_unavailable_percentage = 75
```
bryantbiggs, line 76: nit: lets use 33% since I believe that is our default guidance for update config (cc @allamand)

examples/upstream-with-k8s-addons/outputs.tf
```diff
@@ -0,0 +1,34 @@
...
+  description = "EKS cluster ID"
+  value       = module.eks.cluster_id
+}
+
+output "eks_managed_node_groups" {
+  description = "EKS managed node groups"
+  value       = module.eks.eks_managed_node_groups
+}
+
+output "configure_kubectl" {
```
bryantbiggs, line 26: I think we can remove all but the kubectl output - do we need these other outputs?

Zvikan: > looks good - general question, is `upstream` well known or should we just call it something like `terraform-aws-eks`?

Agree and forgot to call it out, but I first tried to call it `terraform-aws-eks...` but having long name example leads to issues with IRSA/IAM resources names that may be limited to 64 chars :(

Zvikan:
> looks good - general question, is `upstream` well known or should we just call it something like `terraform-aws-eks`?

Agree and forgot to call it out, but I first tried to call it terraform-aws-eks... but having long name example leads to issues with IRSA/IAM resources names that may be limited to 64 chars :(

examples/upstream-with-k8s-addons/outputs.tf
Zvikan: Tried to keep it consistent with other examples outputs, but will remove it and keep the kubectl output

examples/upstream-with-k8s-addons/main.tf
Zvikan: switched to 33

bryantbiggs (approved):

examples/upstream-with-k8s-addons/main.tf
```diff
@@ -0,0 +1,144 @@
...
+  version = "~> 18.0"
+
+  cluster_name                    = local.name
+  cluster_version                 = "1.22"
+  cluster_endpoint_private_access = true
+
+  vpc_id     = module.vpc.vpc_id
+  subnet_ids = module.vpc.private_subnets
+
+  eks_managed_node_group_defaults = {
```
vara-bonthu, line 61: I believe this example should be an upstream module rather than a blueprint.

askulkarni2: @vara-bonthu customers want to see how kubernetes-addons module can be used with upstream.

askulkarni2 (approved):
LGTM!

PR #700, opened 2022-06-28
add helm release addon non-sensitive outputs

FernandoMiguel:
### What does this PR do?

add helm_addons outpus

### Motivation

So practitioners can add outputs to modules

### More

### For Moderators


askulkarni2 (dismissed):
@FernandoMiguel interesting PR. Would you be able to add some more details in the motivation? What is the use-case for being able to consume these outputs? Maybe an code  example would help. 

Also the CI [check](https://github.com/aws-ia/terraform-aws-eks-blueprints/runs/7096811848?check_suite_focus=true) is failing. Running `pre-commit run terraform_docs` should fix that.

FernandoMiguel:
@askulkarni2 I was looking to get the version of the charts installed, and also looking for newer versions that I could upgrade to.
For now, I submitted this just for the helm addon module, and if it accepted I'll open a PR adding outputs to the addons themselves.

I also have a single liner that does what this does, but without descriptions, and also includes the sets and value files, if the team prefers that.
`output "helm_release_external_dns" { value = try({ for k, v in helm_release.external-dns : k => (k != "repository_password" ? v : null) }, null) }`

if you can provide me with further instructions on the pre-commit command, I can try that tomorrow.

FernandoMiguel:
> instead of adding all outputs, would it be possible to only export the outputs that are needed? we could also do a single full export like:
> ```hcl
> output "helm_release" {
>   description = "Attributes of Helm release created"
>   value       = helm_release.addon
> }
> ```

The password attribute is sensitive and would block the all output downstream.

modules/kubernetes-addons/helm-addon/outputs.tf
```diff
@@ -0,0 +1,134 @@
+output "helm_release_addon_name" {
+  description = "Release name."
+  value       = helm_release.addon[0].name
```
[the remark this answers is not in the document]
FernandoMiguel: You are right. I've done it on our module.

modules/kubernetes-addons/helm-addon/outputs.tf
```diff
@@ -0,0 +1,10 @@
+output "helm_release_addon" {
+  description = "Describes the desired status of a chart in a kubernetes cluster."
+  value = try(
+    {
+      for k, v in helm_release.external-dns : k => (
```
bryantbiggs, line 5:
```suggestion
      for k, v in helm_release.addon : k => (
```

FernandoMiguel: That's what I get for copy pasting..
Thanks.

FernandoMiguel:
Pls review.
Thank you.

modules/kubernetes-addons/helm-addon/outputs.tf
```diff
@@ -0,0 +1,10 @@
+output "helm_release_addon" {
+  description = "Describes the desired status of a chart in a kubernetes cluster."
+  value = try(
+    {
+      for k, v in helm_release.addon : k => (
```
bryantbiggs, line 5: I think the value here would be: 
```hcl
value = try({ for k, v in helm_release.addon : k => v if k != "repository_password" }, {})
```

bryantbiggs:
see comment above and it looks like there are CI failures to address (pre-commit)

modules/kubernetes-addons/helm-addon/outputs.tf
FernandoMiguel: Either works in my testing.
I tend to go with not setting explicitly anything, hence null.
But not all providers handle that well.
I'm fine following your existing style, if you prefer.

bryantbiggs (approved):

PR #724, opened 2022-07-01
fix: markdown links, check markdowns on PR

Zvikan:
### What does this PR do?

- fix broken link in docs.
- adjust markdown link action to run on PRs against main only when .md files modified.

### Motivation

- broken main check - https://github.com/aws-ia/terraform-aws-eks-blueprints/runs/7149823004?check_suite_focus=true
- it make sense to check this on PR before changes to main (unless I am missing something here).

### More

### For Moderators

### DIFF
--- a/.github/workflows/markdown-link-check.yml
+++ b/.github/workflows/markdown-link-check.yml
@@ -7,6 +7,12 @@ on:
     paths:
       - "**/*.md"
 
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "**/*.md"
+      
 jobs:
   markdown-link-check:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


askulkarni2 (approved):

.github/workflows/markdown-link-check.yml
```diff
@@ -7,6 +7,12 @@ on:
     paths:
       - "**/*.md"
 
+  pull_request:
+    branches:
+      - main
```
bryantbiggs, line 12: oh - we should chat about this. we had this here but it doesn't work when adding new addons or new pages to docs since its checking whats currently live and not what "would be live" so it fails PRs. 

not sure what the correct route is - maybe we want that and its up to reviewers to know if thats a true failure or a false failure since its getting added in the PR

Zvikan: created https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/725 for testing purposes, looks like the PR check fails as expected but this is coming from a branch created from the same repo, need to test this against a PR coming from external contributor from fork

PR #733, opened 2022-07-05
feat: Add use_kubernetes_provider variable and use_kubectl_provider variable

kahirokunn:
### What does this PR do?

feat: Add use_kubernetes_provider variable and use_kubectl_provider variable

### Motivation

Due to various factors such as network, environment to apply, etc., kubernetes providers may fail to create them.
It may not be possible to work around this by modifying the kubernetes terraform code.
However, the kubectl provider may succeed in creating them.

On the other hand, there may be environments where the kubernetes provider is more successful than the kubectl provider. Therefore, we have made it possible to choose which one to use.

For example, I have tested 100 times to create a service account for the irsa module from Terraform Cloud in the Tokyo region, and it never finishes within the time limit.
I took another approach and tested it by forking the terraform-aws-eks-blueprints repository and adding the timeouts option, but it did not work.
However, I rewrote it in kubectl provider and all is well.

### More

### For Moderators


kahirokunn:
@bryantbiggs
Okay.
I was able to solve this problem with the timeouts option here.
https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/732

Issue #562, opened 2022-05-18
[DOCS] - Add Min IAM Permissions Doc


kcoleman731:
We should provide guidance on min IAM policies that are needed for an admin leverage EKS Blueprints.

Zvikan:
https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/740 created, this will help us generate doc which will be done on a separate PR.
As of now, using my own fork the following list of actions required:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "autoscaling:CreateAutoScalingGroup",
                "autoscaling:CreateOrUpdateTags",
                "autoscaling:DeleteAutoScalingGroup",
                "autoscaling:DeleteLifecycleHook",
                "autoscaling:DeleteTags",
                "autoscaling:DescribeAutoScalingGroups",
                "autoscaling:DescribeLifecycleHooks",
                "autoscaling:DescribeTags",
                "autoscaling:PutLifecycleHook",
                "autoscaling:SetInstanceProtection",
                "autoscaling:UpdateAutoScalingGroup",
                "ec2:AllocateAddress",
                "ec2:AssociateRouteTable",
                "ec2:AttachInternetGateway",
                "ec2:AuthorizeSecurityGroupEgress",
                "ec2:AuthorizeSecurityGroupIngress",
                "ec2:CreateEgressOnlyInternetGateway",
                "ec2:CreateInternetGateway",
                "ec2:CreateLaunchTemplate",
                "ec2:CreateNatGateway",
                "ec2:CreateNetworkAclEntry",
                "ec2:CreateRoute",
                "ec2:CreateRouteTable",
                "ec2:CreateSecurityGroup",
                "ec2:CreateSubnet",
                "ec2:CreateTags",
                "ec2:CreateVpc",
                "ec2:DeleteInternetGateway",
                "ec2:DeleteLaunchTemplate",
                "ec2:DeleteNatGateway",
                "ec2:DeleteNetworkAclEntry",
                "ec2:DeleteRoute",
                "ec2:DeleteRouteTable",
                "ec2:DeleteSecurityGroup",
                "ec2:DeleteSubnet",
                "ec2:DeleteVpc",
                "ec2:DescribeAccountAttributes",
                "ec2:DescribeAddresses",
                "ec2:DescribeAvailabilityZones",
                "ec2:DescribeEgressOnlyInternetGateways",
[...]

PR #740, opened 2022-07-06
Generate IAM Policy with all required Actions

Zvikan:
### What does this PR do?

- Generates IAM policy using [iamlive](https://github.com/iann0036/iamlive) 
  - During e2e parallel workflow, each example will have iamlive running and it will print out the result, once done will upload the result to s3 bucket
  - At the end of the workflow, the workflow will run a python script reading each of the files in the bucket (each policy generated ) and will merge it into a single policy, and will print it out.

### Motivation

- https://github.com/aws-ia/terraform-aws-eks-blueprints/issues/562
- Customers may want to have explicit list of IAM actions required per example or/and for overall using EKS Blueprints.

### More

- [X] Yes, I have tested the PR using my local account setup  (https://github.com/Zvikan/terraform-aws-eks-blueprints/runs/7204477652?check_suite_focus=true)

### For Moderators

### DIFF
--- a/.github/workflows/e2e-parallel-full.yml
+++ b/.github/workflows/e2e-parallel-full.yml
@@ -118,4 +118,31 @@ jobs:
           kill ${{ env.iamlive_pid }}
           while $(kill -0 ${{ env.iamlive_pid }} 2>/dev/null); do sleep 1; done;
           cat ${HOME}/policy.json
-          aws s3 cp ${HOME}/policy.json s3://terraform-eks-blueprints-iam-policies-examples/${{ matrix.example_path }}.json
+          aws s3 cp ${HOME}/policy.json s3://eks-blueprints-iam-policies/${{ matrix.example_path }}.json
+
+    post_deploy:
+      if: ${{ always() }}
+      needs: [deploy]
+      permissions:
+        id-token: write
+        contents: read
+      name: Merge Policies and Print Final IAM Policy
+      runs-on: ubuntu-latest
+      steps:
+        # Be careful not to change this to explicit checkout from PR ref/code, as below we run a python code that may change from the PR code.
+        - name: Checkout
+          uses: actions/checkout@v3
+
+        - name: Configure AWS credentials from Test account
+          uses: aws-actions/configure-aws-credentials@v1
+          with:
+            role-to-assume: ${{ secrets.ROLE_TO_ASSUME }}
+            aws-region: us-west-2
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-parallel-full.yml
```diff
@@ -57,6 +57,19 @@ jobs:
           role-duration-seconds: 3600
           role-session-name: GithubActions-Session
 
+      - name: Iamlive Setup
+        run: |
+          #!/bin/bash
+          set -e
```
bryantbiggs, line 63: nit: Could combine into `set -eox pipefail`

.github/workflows/e2e-parallel-full.yml
```diff
@@ -57,6 +57,19 @@ jobs:
...
+          set -o pipefail
+          set -x
+          export IAMLIVE_VERSION=v0.48.0
```
bryantbiggs, line 66: nit and personal opinion: what if we put this up top in the global env like https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/.github/workflows/pre-commit.yaml#L9

askulkarni2 (commented):
Nice! Can we add a doc page that shows the policy for a few key examples?

.github/workflows/e2e-parallel-full.yml
```diff
@@ -86,4 +107,42 @@ jobs:
...
+          cat ${HOME}/policy.json
           exit 1
+
+      - name: Iamlive Print & Upload Policy
+        if: ${{ always() }}
+        run: |
+          kill ${{ env.iamlive_pid }}
+          while $(kill -0 ${{ env.iamlive_pid }} 2>/dev/null); do sleep 1; done;
+          cat ${HOME}/policy.json
+          aws s3 cp ${HOME}/policy.json s3://eks-blueprints-iam-policies/${{ matrix.example_path }}.json
```
bryantbiggs, line 121: 1. We should make the bucket name either a workflow global env variable or a github secret to ensure its defined in one place correctly (is it `terraform-eks-blueprints-iam-policies-examples` or `eks-blueprints-iam-policies` ?)
2. do we need to add in a git ref or something so that PRs aren't clobbering one another? adding something like `/${{ github.ref }}/` into the prefix?
3. If we do that, do we need/want to set a lifecycle policy on these

askulkarni2 (approved):
LGTM!

PR #745, opened 2022-07-07
hotfix: remove grafana secret from spark example until fixed

Zvikan:
### What does this PR do?

- Removing grafana secret manager password from spark example. 

### Motivation

- Post https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/735 there are pre-commit issues and failures with tf plan and apply to the spark example due to the implementation to the secret manager handling. This is currently blocking other on-going work. 
- https://github.com/aws-ia/terraform-aws-eks-blueprints/runs/7239480579?check_suite_focus=true
- https://github.com/aws-ia/terraform-aws-eks-blueprints/runs/7239282268?check_suite_focus=true
- Tried to hotfix it by changing code to use implicit dependency between resource to `grafana_admin_password_secret_name` - lead to the following error:
```
 Error: Invalid count argument
│
│   on ../../../modules/kubernetes-addons/grafana/data.tf line 3, in data "aws_secretsmanager_secret" "admin_password":
│    3:   count = var.grafana_admin_password_secret_name == "" ? 0 : 1
│
│ The "count" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created. To work around this, use the -target argument to first
│ apply only the resources that the count depends on.
``` 

Removing for now until fixed, also enabled status checkes to be required before merging.

### More

### For Moderators

### DIFF
--- a/examples/analytics/spark-k8s-operator/main.tf
+++ b/examples/analytics/spark-k8s-operator/main.tf
@@ -202,7 +202,6 @@ module "eks_blueprints_kubernetes_addons" {
   # Open Source Grafana Add-on
   #---------------------------------------------------------------
   enable_grafana                     = true
-  grafana_admin_password_secret_name = ""
 
   tags = local.tags
 }
--- a/examples/analytics/spark-k8s-operator/main.tf
+++ b/examples/analytics/spark-k8s-operator/main.tf
@@ -41,7 +41,6 @@ locals {
   vpc_cidr                           = "10.0.0.0/16"
   azs                                = slice(data.aws_availability_zones.available.names, 0, 3)
   s3_prefix                          = "logs/"
-  grafana_admin_password_secret_name = "grafana"
 
   tags = {
     Blueprint  = local.name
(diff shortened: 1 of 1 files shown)


vara-bonthu:
Yes, I agree with the above points. I handled this better way base don the feedback in this new PR #746 

- This allows users to pass any secret using helm chart set_sensitive values
- This also shows an example of how to create a new password using SecretsManager and pass that through custom values

I think we need to apply the similar approach to ArgoCD as well. Let me know your feedback on the new PR

Zvikan:
> Yes, I agree with the above points. I handled this better way based on the feedback in this new PR #746
> 
> * This allows users to pass any secret using helm chart set_sensitive values
> * This also shows an example of how to create a new password using SecretsManager and pass that through custom values
> 
> I think we need to apply the similar approach to ArgoCD as well. Let me know your feedback on the new PR

Thanks Vara for the PR, let's close this one once the other PR is merged.

bryantbiggs:
Closed in #746

PR #799, opened 2022-07-24
Self-managed node group - security group fix

awsitcloudpro:
### What does this PR do?

- Added logic to automatically attach EKS cluster security group to self-managed node groups, and ensure only the cluster security group has the tag `kubernetes.io/cluster/[cluster-name] = owned`
- Updated instance types in Windows node group example
- Updated docs with a note about security groups for self-managed node groups
 
### Motivation

- If both self-managed node groups and managed node groups are created in a single cluster, this module's default configuration results in self-managed node groups that have network connectivity issues with the managed node groups. This creates issues - for example, pods running on self-managed node groups not being able to resolve DNS names due to inability to connect to CoreDNS pods running on managed node groups. Attaching the EKS cluster security group to self-managed node groups takes care of resolving all these network connectivity issues. However, the [default configuration](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/58285f37d1dfaacbf99055ed87f1be19cc39f5ad/node_groups.tf#L160) of the community eks module adds the tag `kubernetes.io/cluster/[cluster-name] = owned` to the self-managed node security group by default. This [prevents](https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1181) AWS Load Balancer Controller to provision ingress-related rules in the cluster security group, resulting in the controller being unusable for ingresses backed by pods running on self-managed node groups. The [troubleshooting guide](https://aws.amazon.com/premiumsupport/knowledge-center/load-balancer-troubleshoot-creating) for ALB controller mentions the need to tag exactly one security group with the cluster "owned" tag.
- Instance type m5.large used by the windows node group example is not optimal for Windows workloads. Multiple simultaneously running Windows pods typically starve for resources on that instance type. Also, managed node groups / spot instance types should use similar configuration when using Cluster Autoscaler for autoscaling.

### More

- [x] Yes, I have tested the PR using my local account setup  (Tested with multiple clusters with only managed, only self-managed, and mixed node-groups, with and without Windows workloads.)

**Note**: Not all the PRs require examples and docs except a new pattern or add-on added.

### For Moderators


docs/node-groups.md
```diff
@@ -744,3 +744,7 @@ For example, if you are enabling the `metrics-server` Kubernetes add-on (Helm ch
     ]
   }
 ```
+
+### Security groups attached to self-managed node groups
+
+EKS automatically creates a [cluster security group](https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html) and attaches it to any managed node groups in the cluster. This framework attaches the EKS cluster security group along with an additional security group to self-managed node groups, so that they can communicate with any EKS managed node groups that may be present in the cluster. As recommended in the AWS Load Balancer Controller [troubleshooting guide](https://aws.amazon.com/premiumsupport/knowledge-center/load-balancer-troubleshoot-creating/), the framework ensures that only one of the security groups - the EKS cluster security group - is tagged with tag `kubernetes.io/cluster/[cluster-name] = "owned"`, so that AWS Load Balancer Controller can create necessary rules for ingresses in the EKS cluster security group.
```
bryantbiggs, line 750: the cluster primary security group is only attached to nodes in the following scenarios:
1. Fargate profiles 
2. The default launch template is used for EKS managed node groups

examples/node-groups/windows-node-groups/main.tf
```diff
@@ -51,7 +51,7 @@ module "eks_blueprints" {
     mng_spot_medium = {
       node_group_name = "mng-spot-med"
       capacity_type   = "SPOT"
-      instance_types  = ["t3.large", "t3.xlarge"]
+      instance_types  = ["t3.large", "m6i.large", "m6a.large", "m5n.large"]
```
bryantbiggs, line 54: what is the reason for the instance type changes?

locals.tf
```diff
@@ -19,7 +19,14 @@ locals {
   public_subnet_ids  = var.public_subnet_ids
 
   enable_workers            = length(var.self_managed_node_groups) > 0 || length(var.managed_node_groups) > 0 ? true : false
-  worker_security_group_ids = local.enable_workers ? compact(flatten([[module.aws_eks.node_security_group_id], var.worker_additional_security_group_ids])) : []
+  worker_security_group_ids = local.enable_workers ? compact(flatten([[module.aws_eks.cluster_primary_security_group_id, module.aws_eks.node_security_group_id], var.worker_additional_security_group_ids])) : []
```
bryantbiggs, line 22: this looks massively complex and brittle. Instead of forcing this upon users, I think educating users of the various options and "gotchas" to be aware of would be more beneficial

docs/node-groups.md
awsitcloudpro: Looks like there is some confusion here. This README change is a description of *this particular change in the PR*. After the PR is merged, the README text will be valid - i.e. this framework (the Terraform AWS EKS Blueprints module) will attach EKS cluster security group to self-managed node groups.

examples/node-groups/windows-node-groups/main.tf
awsitcloudpro: Refer to [Cluster Autoscaler best practices](https://aws.github.io/aws-eks-best-practices/cluster-autoscaling/#configuring-your-node-groups). If instance types with different CPU and memory capacity are used for horizontal autoscaling, resources may get wasted.

bryantbiggs:
Closed in #876

PR #813, opened 2022-07-28
cert manager default behavior is to keep certificate after ingress/c…

avihaycloudinary:
### What does this PR do?

I add this pr since we faced rate limit during developing in letsencrypt cert manager, u can read more about it [here](https://letsencrypt.org/docs/rate-limits/) 
It took me time to find out the reason and how to solve it, and I was not the only one [looking for a solution](https://github.com/cert-manager/cert-manager/issues/3298#issue-704410330).
So to make it simpler to other **I added a designated parameter to make it clear**, as well I changed the blueprint default behavior to be as cert-manager default behavior. 
cert manager default behavior is to keep certificate after ingress/certificate deletion. currently this behavior is override and u might hit letsencrypt rate limit (weekly) so it is long wait time. 
see discussion https://github.com/cert-manager/cert-manager/issues/3298#issuecomment-1030815199

### More

### For Moderators

### DIFF
--- a/modules/kubernetes-addons/cert-manager/values.yaml
+++ b/modules/kubernetes-addons/cert-manager/values.yaml
@@ -1,5 +1,5 @@
 extraArgs:
-  - --enable-certificate-owner-ref=true
+  - --enable-certificate-owner-ref=${delete_permanently}
 
 installCRDs: true
 
--- a/modules/kubernetes-addons/README.md
+++ b/modules/kubernetes-addons/README.md
@@ -111,6 +111,7 @@
 | <a name="input_aws_privateca_acmca_arn"></a> [aws\_privateca\_acmca\_arn](#input\_aws\_privateca\_acmca\_arn) | ARN of AWS ACM PCA | `string` | `""` | no |
 | <a name="input_aws_privateca_issuer_helm_config"></a> [aws\_privateca\_issuer\_helm\_config](#input\_aws\_privateca\_issuer\_helm\_config) | PCA Issuer Helm Chart config | `any` | `{}` | no |
 | <a name="input_aws_privateca_issuer_irsa_policies"></a> [aws\_privateca\_issuer\_irsa\_policies](#input\_aws\_privateca\_issuer\_irsa\_policies) | IAM policy ARNs for AWS ACM PCA IRSA | `list(string)` | `[]` | no |
+| <a name="input_cert_manager_allow_reuse_cert_after_delete"></a> [cert\_manager\_allow\_reuse\_cert\_after\_delete](#input\_cert\_manager\_allow\_reuse\_cert\_after\_delete) | To avoid recreation of certificate when service is being delete and recreate.(can assist with letsencrypt rate limit) | `bool` | `true` | no |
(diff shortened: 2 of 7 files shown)


vara-bonthu (approved):
Thanks for the PR, @avihaycloudinary 👍 

This use case seems to be valid. LGTM! 
Could you please post the test results as well, if any?

avihaycloudinary:
> Thanks for the PR, @avihaycloudinary 👍
> 
> This use case seems to be valid. LGTM! Could you please post the test results as well, if any?

Hi,
Thanks for a quick reply, 🙏 
I'm on a holiday, I will publish the test when be back.
👍

askulkarni2 (commented):
Thanks for the PR, looks good. Ready to merge after testing is validated.

avihaycloudinary:
> Thanks for the PR, looks good. Ready to merge after testing is validated.

Testing Scenarios:

1) **update an existing cluster with new definition** 
```
  # module.eks_cluster_preprod.module.eks_blueprints_kubernetes_addons.module.cert_manager[0].module.helm_addon.helm_release.addon[0] will be updated in-place
  ~ resource "helm_release" "addon" {
        id                         = "cert-manager"
        name                       = "cert-manager"
      ~ values                     = [
          - <<-EOT
                extraArgs:
                  - --enable-certificate-owner-ref=true
                
                installCRDs: true
                
                securityContext:
                  enabled: true
                  fsGroup: 1001
            EOT,
          + <<-EOT
                extraArgs:
                  - --enable-certificate-owner-ref=false
                
                installCRDs: true
                
                securityContext:
                  enabled: true
                  fsGroup: 1001
            EOT,
        ]
        # (28 unchanged attributes hidden)


        # (3 unchanged blocks hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
╷
```

2) **recreated certificate object (crd) cluster cert manager definiton enable-certificate-owner-ref=true** 
certificate was created and stored as secret (took over 1 minute to be ready)
deleting certificate object, secret was deleted 
recreate certificate (took over 1 minute to be ready)

Total API calls to lets encrypt is 2

3) **recreated certificate object (crd) cluster cert manager definiton enable-certificate-owner-ref=false** 
certificate was created and stored as secret (took over 1 minute to be ready)
deleting certificate object, secret was **not** deleted 
recreate certificate (secret available immediately)

Total API calls to lets encrypt is 1

modules/kubernetes-addons/cert-manager/values.yaml
```diff
@@ -1,5 +1,5 @@
 extraArgs:
-  - --enable-certificate-owner-ref=true
```
bryantbiggs, line 2: while I agree that the change is valid - I think the better solution might be to just simply remove the extaArgs here and let users pass in the args that suit their use case. 

I think I would prefer that we simply remove the `extraArgs` from the default values file here and leave this as an implementation detail for users. In fact, we can remove the default values file entirely and just set `installCRDs: true` in `set_values` like https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/modules/kubernetes-addons/aws-coredns/main.tf#L59-L64

@vara-bonthu  / @askulkarni2  thoughts?

PR #821, opened 2022-08-02
Add option to use existing CloudWatch log group for aws-for-fluentbit

kbrockhoff:
### What does this PR do?

Added flag to supply existing log group for use by was-for-fluentbit. Defaults to current behavior so no breaking change.

### Motivation

Need for this was discovered because of requirement to send logs to 3 different log groups depending on who needs to look at them. Wanted to use the same KMS key to encrypt all 3 log groups. Originally tried to create KMS CMK and 2 log groups outside blueprints and supply KMS key to blueprints module. However, this resulted in Terraform error because it cannot determine the count for the fluent bit KMS key resource. To fix, would need a flag on whether a KMS key should be created or not but this would be a breaking change. So went route of supplying log group and a corresponding flag which was implemented without a breaking change. Also make the external code clearer because all 3 log groups are specified in the same tf file.

### More

- [X ] Yes, I have tested the PR using my local account setup  (Provide any test evidence report under Additional Notes)
- [ X] Yes, I ran `pre-commit run -a` with this PR

### For Moderators


vara-bonthu (commented):
Thanks for the PR @kbrockhoff  👍🏼 

Overall PR Looks good to me. We may not need a brand new example for this change. You can add the usage details in `docs/add-ons/aws-for-fluent-bit.md`

kbrockhoff:
> Thanks for the PR @kbrockhoff 👍🏼
> 
> Overall PR Looks good to me. We may not need a brand new example for this change. You can add the usage details in `docs/add-ons/aws-for-fluent-bit.md`

I removed the example I had added and clarified the wording in the documentation I added.

docs/add-ons/aws-for-fluent-bit.md
```diff
@@ -54,3 +54,17 @@ awsForFluentBit = {
   logGroupName = "<log_group_name>"
 }
 ```
+
+### Externally-Created CloudWatch Log Group(s)
+
+If the CloudWatch log group FluentBit puts logs to is required to be encrypted by an existing KMS
+custoerm-managed key, then the CloudWatch log group needs to be created external to the
```
askulkarni2, line 61: nit: custoerm-managed, change to "customer managed"

askulkarni2 (approved):
Thanks for the PR! LGTM!

PR #834, opened 2022-08-09
feat: Allow setting `set_values` through `helm_config` for AWS load balancer controller addon

bryantbiggs:
### What does this PR do?
- Allow setting `set_values` through `helm_config` for AWS load balancer controller addon
- Changed the Fargate example (where the above change is demonstrated/tested) to deploy the sample app by default

### Motivation
- In order to set additional values, currently users would need to duplicate the values file which is not a great user experience. Instead, allowing users to pass in additional values through `set_values` allows for the default values file to stay as is as the minimum required values, and users can pass in any additional through the `var.helm_config.set_values`. The motivation for being able to set additional values - to able to set the pod disruption budget on the AWS LBC for better HA (discovered during testing for cluster upgrade guidance)
- Deploying the sample app by default for the Fargate example is just to simplify things

### More

### For Moderators


examples/fargate-serverless/main.tf
```diff
@@ -137,12 +131,15 @@ module "eks_blueprints_kubernetes_addons" {
 
   enable_aws_load_balancer_controller = true
   aws_load_balancer_controller_helm_config = {
-    values = [
-      <<-EOT
-      clusterName: ${module.eks_blueprints.eks_cluster_id}
-      region: ${local.region}
-      vpcId: ${module.vpc.vpc_id}
```
bryantbiggs, line 144: we previously had to replicate the values file here in order to set the `vpcId` for running the AWS LBC on Fargate. now we can set just the additional values we need to provide to the AWS LBC addon

examples/fargate-serverless/main.tf
```diff
@@ -268,27 +265,18 @@ module "vpc" {
...
-  ]
 }
 
 resource "kubectl_manifest" "sample_app" {
-  count     = var.deploy_sample_app ? length(data.kubectl_path_documents.sample_app.documents) : 0
-  yaml_body = element(data.kubectl_path_documents.sample_app.documents, count.index)
-  depends_on = [
-    kubernetes_namespace_v1.sample_app[0]
-  ]
+  yaml_body = templatefile("sample_app.yaml", {
```
bryantbiggs, line 278: flattened to one multi-resource yaml file with templating so that we can map dependencies (by passing namespace from namespace resource)

modules/kubernetes-addons/aws-load-balancer-controller/locals.tf
```diff
@@ -6,7 +6,7 @@ locals {
     name        = local.name
     chart       = local.name
     repository  = "https://aws.github.io/eks-charts"
-    version     = "1.4.2"
+    version     = "1.4.3"
```
FernandoMiguel, line 9: 1.4.4 is out today

Zvikan (approved):
LGTM, we should consider applying the same change to all the other addons following the same motivation mentioned here.

PR #852, opened 2022-08-13
add fail-fast for name length restrictions

kbrockhoff:
### What does this PR do?

Adds a length restrictions on cluster_name and add-on service account name. It also shortens the service account name of two existing add-ons. The goal being to ensure generated IAM role names do not exceed the 64 char maximum.

### Motivation

We use hierarchical-based naming scheme with all our cloud resources. Names tend to get long. It is a time waster to trigger cluster creation which then fails after 20-30 minutes because an IAM Role name exceeded its maximum length. Also, with current length of several generated IAM Role names, it is does not leave enough characters available for cluster name that we can use are standard naming scheme.

### More

- [X ] Yes, I have tested the PR using my local account setup (Provide any test evidence report under Additional Notes)
- [X ] Yes, I ran `pre-commit run -a` with this PR

### For Moderators


modules/kubernetes-addons/aws-node-termination-handler/locals.tf
```diff
@@ -1,6 +1,6 @@
 locals {
   namespace            = "kube-system"
-  name                 = "aws-node-termination-handler"
```
bryantbiggs, line 3: these will fail since the chart name is something we don't control but set by the chart. the release name can change, the chart name cannot

kbrockhoff: Changed naming here and verified it was all created and working.

askulkarni2:
@kbrockhoff can you please open an issue for this? Also if the IAM role name lengths are an issue should we try to `substr()` [here](https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/860fd783c417adf3a9d5a58329956a67a0b062c7/modules/irsa/main.tf#L27) instead? This change of SA names might break the state of folks using those addons. @bryantbiggs @vara-bonthu @Zvikan thoughts?

Issue #854, opened 2022-08-15
Unable to attach alb to nodes autoscaling group


olibenni:
## Description
Using aws_autoscaling_attachment to attach alb target group arn with out autoscaling group for the nodes in the eks
only works on every other apply. The apply is successful but the attachment is removed on every other apply.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## Versions
- Module version [Required]: 4.6.2

## Reproduction Code [Required]
``` hcl
module "eks_blueprints" {
  source = "github.com/aws-ia/terraform-aws-eks-blueprints?ref=v4.6.2"

  cluster_version    = "1.21"
  vpc_id             = var.vpc_id
  private_subnet_ids = var.subnet_ids
  cluster_name       = "somecluster-eks"

  self_managed_node_groups = {
    self_mg5 = {
      node_group_name    = "self_mg5"
      instance_type      = "m5.large"
      launch_template_os = "amazonlinux2eks"
      subnet_ids         = var.worker_subnet_ids
      public_ip          = true
      # This is an invalid argument at the moment
      # target_group_arns  = [var.aws_lb_target_group_arn]
  }
}

resource "aws_autoscaling_attachment" "asg_attachment_bar" {
  autoscaling_group_name = module.eks_blueprints.self_managed_node_group_autoscaling_groups[0]
  alb_target_group_arn   = var.aws_lb_target_group_arn
}
```

## Expected behavior
Expect new nodes need be be part of our target group
The autoscaling group in the eks cluster needs to be aware of the target group so that new nodes are part of it

## Actual behavior
When we first apply, the attachment is made and things work fine.
**But in our subsequent apply, the eks_blueprint notices an attachment that isn't defined in the blueprint so it removes it.**

askulkarni2:
@olibenni out of curiosity, what is your use-case to attach an ALB directly to the ASG this way?

olibenni:
Good question, I wanted to add a loadbalancer for the ArgoCD server, exposing the ArgoCD web ui.
Open to alternatives.

askulkarni2:
[...]

PR #855, opened 2022-08-15
Pass target_group_arns to self managed nodes autoscaling group

olibenni:
### What does this PR do?
Pass target_group_arns to self managed nodes autoscaling group
See issue: https://github.com/aws-ia/terraform-aws-eks-blueprints/issues/854

### Motivation
When adding an alb for our nodes there was no way of passing down the target_group_arn to the nodes
This can be circumvented by using an autoscaling attachment but in every other apply the eks_blueprint notices an attachment that isn't defined in the blueprint so it removes it.
See issue: https://github.com/aws-ia/terraform-aws-eks-blueprints/issues/854

### More

### For Moderators

### DIFF
--- a/modules/aws-eks-self-managed-node-groups/main.tf
+++ b/modules/aws-eks-self-managed-node-groups/main.tf
@@ -6,6 +6,7 @@ resource "aws_autoscaling_group" "self_managed_ng" {
   vpc_zone_identifier = length(local.self_managed_node_group["subnet_ids"]) == 0 ? (local.self_managed_node_group["subnet_type"] == "public" ? var.context.public_subnet_ids : var.context.private_subnet_ids) : local.self_managed_node_group["subnet_ids"]
 
   capacity_rebalance = local.self_managed_node_group["capacity_rebalance"]
+  target_group_arns  = local.self_managed_node_group["target_group_arns"]
 
   dynamic "launch_template" {
     for_each = !local.needs_mixed_instances_policy ? [1] : []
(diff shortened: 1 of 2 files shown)


vara-bonthu (approved):
LGTM! Thanks @olibenni 👍🏼

vara-bonthu:
@olibenni Let's make sure if all the checks are passed.

bryantbiggs:
FYI - this won't work as intended unless the argument is added to the underlying autoscaling group resource

modules/aws-eks-self-managed-node-groups/main.tf
```diff
@@ -6,6 +6,7 @@ resource "aws_autoscaling_group" "self_managed_ng" {
   vpc_zone_identifier = length(local.self_managed_node_group["subnet_ids"]) == 0 ? (local.self_managed_node_group["subnet_type"] == "public" ? var.context.public_subnet_ids : var.context.private_subnet_ids) : local.self_managed_node_group["subnet_ids"]
 
   capacity_rebalance = local.self_managed_node_group["capacity_rebalance"]
+  target_group_arns  = local.self_managed_node_group["target_group_arns"]
```
bryantbiggs, line 9: Instead of defaulting to an empty list, we should default to `null` - for any users that are using the `aws_autoscaling_attachment` resource, if we default to an empty list this will cause conflicts for them whereas `null` should not (I believe).  With the proposed change below, you can remove the change to the `locals.tf` file

See more info in the `Note` section at the top of https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/autoscaling_group

```suggestion
  target_group_arns  = try(local.self_managed_node_group["target_group_arns"], null)
```

olibenni: Thanks, updated

askulkarni2 (approved):
Thanks for the PR @olibenni!

olibenni:
My pleasure 🤝

Issue #494, opened 2022-05-01
[FEATURE] Add some easy to see cautionary notes and helpful hints around KMS usage in the README & examples


spkane:
#### Is your feature request related to a problem? Please describe

After helping out with this https://github.com/aws-ia/terraform-aws-eks-blueprints/issues/485 and having had a similar issue early on here https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/414

It might be good to make sure that people using this module are aware of the basics of KMS and how they can protect themselves from unmanageable keys. It really is a pain when this happens, especially when you aren't really familiar with KMS and all of a sudden find yourself with a terraform stack that can't be easily destroyed and replaced.

#### Describe the solution you'd like

Add some very easy-to-discover documentation in the README and default examples for new users with a few warnings, and explanations of best practices, and bad outcomes.

bryantbiggs:
Thank you for creating this - I agree, theres a bit of work to do around KMS; possibly even externalizing it and instead showing example usage from an `example/` rather than create within the root module.

vrodriguezcl:
Will we see an any boolean option if we want to use kms or not ? I mean, something like: enable_kms = true/false in our tf file ?. @bryantbiggs

bryantbiggs:
This will be resolved in the upcoming v5 release

bryantbiggs:
closed in #857

Issue #485, opened 2022-04-27
[QUESTION] Cannot Delete KMS after Creation with Terraform


vrodriguezcl:
#### Please describe your question here:

Recently I was playing with this project, and after the cluster creation, everything was Ok. But, in the time to destroy the created infrastructure using Terraform, one resource specifically cannot be deleted. And I tried many times to delete it an I cannot, I got this error when I tried to delete it from the AWS Web Console, and using **awscli** I cannot get any output.

AccessDeniedException - User: arn:aws:iam::111111111111:user/kube-administrator is not authorized to perform: kms:DisableKey on resource: arn:aws:kms:us-east-1:111111111111:key/be374906-a2d3-21ad-8ef5-25b2cb8048a0 because no resource-based policy allows the kms:DisableKey action

To continue testing this project, I added the **kms module** in my **main.tf** but, changing the values of the alias name (**alias/custom-name**), it cannot be shown in the **tf plan**.

If I need to use kms module is required referencing the URL from GitHub as the same way as the addons example show us?.

Please, help me of how to delete this **KMS** or give me the steps to use a custom KMS name.

I'm using **Terraform Cloud** and **1.1.9** as Terraform Binary version.

Just to know, how can I add a default alias/custom-name to my KMS to avoid that terraform validates that this currently one exists and create a new one?. Or, exists any walkthrough to avoid the kms creation ?. Something like: **aws_kms_enabled = true/false** and put this parameter in my **main.tf** file ?. This module is a must for the eks cluster creation ?. Or we can skip this ?

#### Additional context:
            "Sid": "Allow access for Key Administrators",
            "Effect": "Allow",
            "Principal": {
                "AWS": [
                    "**_AROAYWHCTDYSNS3AU5YK4_**",
                    "arn:aws:iam::111111111111:user/kube-administrator"
                ]

#### More:

- [X] Yes, I have checked the repo for existing issues before raising this question

spkane:
You have to look at the key policy. By default only two roles are allowed to delete the KMS aliases or keys (have `kms:*`). KMS keys are very sensitive and highly protected.

If you delete those roles by hand, etc. it is possible to make a KMS key unmanageable, in which case you will need help from AWS support.

https://docs.aws.amazon.com/kms/latest/developerguide/key-policy-default.html

This is why there is now a new variable that lets you add additional ARNs that you would like to have full access to the KMS key created in this module.

[...]

PR #858, opened 2022-08-15
refactor: Replace local Fargate profile sub-module with EKS module implementation

### LINKED ISSUE #751: Replace Fargate profile module
#### Is your feature request related to a problem? Please describe
- Replace the local implementation of Fargate profile with the implementation provided by the `terraform-aws-eks` module https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/modules/fargate-profile

#### Describe the solution you'd like
- The local Fargate profile is replaced with https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/modules/fargate-profile

#### Additional context
- Changes that users need to make when upgrading from v4.x to v5.x need to be added to the upgrade guide

bryantbiggs:
### What does this PR do?
- Replace local Fargate profile sub-module with EKS module implementation
- See upgrade guide for additional details

### Motivation
- Closes #751

### More

#### How this was tested and validated

1. On `main`, cd  to `examples/fargate-serverless`
2. `terraform init -upgrade=true`
3. `terraform apply` - enter `yes` at prompt
4. TODO - need to look into why sample app doesn't deploy properly, will need to `terraform apply` a second time on failure of sample app deployment (exists on `main` currently as well)
5. `terraform apply -refresh-only` to clean up plan diff
6. `git checkout refactor/replace-fargate-profile-module`
7. Copy terraform state move's provided in upgrade guide to migrate Fargate profiles (see bottom of guide) - perform move commands
8. Update example for backwards compatibility changes:
	- Add 
	```hcl
	  fargate_profile_defaults = {
	      # To maintain backwards compatibility w/ v4.x
	      iam_role_use_name_prefix   = false
	      iam_role_attach_cni_policy = false
	    }
	```
	- Add `iam_role_name = "${local.name}-<PROFILE-NAME>"` to each profile definition block
	- See below to view/make changes described above for backwards compatibility 
	```hcl
	module "eks_blueprints" {
	    source = "../.."
	  
	    cluster_name    = local.name
	    cluster_version = "1.22"
	  
	    vpc_id             = module.vpc.vpc_id
	    private_subnet_ids = module.vpc.private_subnets
	  
	    # https://github.com/aws-ia/terraform-aws-eks-blueprints/issues/485
	    # https://github.com/aws-ia/terraform-aws-eks-blueprints/issues/494
	    cluster_kms_key_additional_admin_arns = [data.aws_caller_identity.current.arn]
	  
	    fargate_profile_defaults = {
	      # To maintain backwards compatibility w/ v4.x
	      iam_role_use_name_prefix   = false
	      iam_role_attach_cni_policy = false
	    }
	  
	    fargate_profiles = {
	      # Providing compute for default namespace
	      default = {
	        name          = "default"
	        iam_role_name = "${local.name}-default"
	        selectors = [
	          {
	            namespace = "default"
	          }
	        ]
	      }
	      # Providing compute for kube-system namespace where core addons reside
	      kube_system = {
	        name          = "kube-system"
	        iam_role_name = "${local.name}-kube-system"
	        selectors = [
	          {
	            namespace = "kube-system"
	          }
	        ]
	      }
	  
	      alb_sample_app = {
	        name          = "alb-sample-app"
	        iam_role_name = "${local.name}-alb-sample-app"
	        selectors = [
	          {
	            namespace = local.sample_app_namespace
	          }
	        ]
	      }
	    }
	  
	    tags = local.tags
	  }
	```
9. Comment out the two `depends_on` blocks - everything is deployed so we don't need this explicit coordination, however having them in place will force additional changes during the migration that we want to avoid (added noise). See note added to upgrade guide (at top)
10. `terraform plan` should show a relatively clean plan save for some tag updates, IAM sid name change, IAM description change, and the removal of the CloudWatch logs policy attachment


examples/complete-kubernetes-addons/main.tf
```diff
@@ -108,18 +108,19 @@ module "eks_blueprints" {
...
         {
           namespace = "default"
-          k8s_labels = {
+          labels = {
             Environment = "preprod"
             Zone        = "dev"
             env         = "fargate"
           }
-      }]
-      subnet_ids = module.vpc.private_subnets
```
Zvikan, line 121: I assume you removed it before the default behavior is to use the same private subnets as provided to the cluster, do we think it may be beneficial to show how to use specific subnets?

Zvikan (approved):
LGTM. one minor comment.
We should probably also fix anything under `docs/` to align with v5 changes, should we do it on this PR?

examples/complete-kubernetes-addons/main.tf
bryantbiggs: I would say - it depends. what do we want to show here under this example thats called "complete-kubernetes-addons"?

bryantbiggs:
@Zvikan for docs, in https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/refactor/replace-self-managed-node-group-module I am removing the `docs/node-groups.md` - are there other specific areas that should be updated for this change?

I think in general, the docs need to be updated just glancing through (outdated module versions). Maybe we can do that on the v5 branch before merging to `main`

examples/complete-kubernetes-addons/main.tf
Zvikan: > I would say - it depends. what do we want to show here under this example thats called "complete-kubernetes-addons"?

Fair, maybe we should consider adding this in the appropriate [example](https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/cbc31b0bbb2c859681a2e456f13850ec5bad260b/examples/fargate-serverless/main.tf#L61-L91) instead, WDYT?

bryantbiggs: I think there are two avenues here that we are mixing
1. How to use the `terraform-aws-eks` module
2. How to achieve an outcome

For one, we can point to the appropriate location that demonstrates all the different ways the underlying module can be configured https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/fargate_profile/main.tf

Point 2 is what I think we are mostly concerned with here (IMO). If someone looks at the `serverless-fargate` example and sees the different way of specifying subnets - does that mean thats how they have to be configured to create a serverless cluster running Fargate?

Issue #810, opened 2022-07-28
[ADDON] - Apache Airflow for EKS


askulkarni2:
- Kubernetes Add-on for Apache Airflow deployment in EKS
- Example pattern for deploying production grade Apache Airflow deployment in EKS with all the best practices
- Documentation for Airflow usage with few examples

bryantbiggs:
closed in #878

PR #878, opened 2022-08-22
Self-managed Apache Airflow on EKS

vara-bonthu:
### What does this PR do?

- This PR deploys Apache Airflow Add-on on EKS
- Production ready deployment example with instructions
- Github Docs page

Closes the issue: #810 

### Motivation

Running Apache Airflow on Kubernetes is the common pattern for Data teams. This example will help customers to build production ready Apache Airflow on EKS.

This is also a request from few AWS customers

### More

### For Moderators


examples/analytics/airflow-on-eks/main.tf
```diff
@@ -0,0 +1,542 @@
...
+        Zone          = "test"
+        WorkerType    = "ON_DEMAND"
+        NodeGroupType = "core"
+      }
+
+      # See this doc node-template tags https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-can-i-scale-a-node-group-to-0
+      additional_tags = {
+        Name                                                             = "core-node-grp"
+        subnet_type                                                      = "private"
+        "k8s.io/cluster-autoscaler/node-template/label/arch"             = "x86"
```
bryantbiggs, line 141: I didn't see `cluster-autoscaler` addon enabled - do we want to enable that here, or remove these tags?

examples/analytics/airflow-on-eks/main.tf
```diff
@@ -0,0 +1,542 @@
...
+      value = 1
+    },
+    {
+      name  = "client_encoding"
+      value = "utf8"
+    }
+  ]
+
+  tags = local.tags
+  db_option_group_tags = {
```
bryantbiggs, line 264: I assume we could probably remove these two tags since they are not related to the outcome here - avoid any confusion for users

examples/analytics/airflow-on-eks/main.tf
```diff
@@ -0,0 +1,542 @@
...
+  db_parameter_group_tags = {
+    "Sensitive" = "low"
+  }
+}
+
+#---------------------------------------------------------------
+# S3 bucket for Airflow Logs
+#---------------------------------------------------------------
+#tfsec:ignore:aws-s3-enable-bucket-logging tfsec:ignore:aws-s3-enable-versioning
+resource "aws_s3_bucket" "this" {
```
bryantbiggs, line 276: FYI - Theres a module for this that makes examples a bit easier as well with the `force_destroy` option plus canned bucket policies https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/9b9686dcb8cc7a2a3717b4e582de39e6d60fabdf/examples/managed-workflow-apache-airflow/main.tf#L238

examples/analytics/airflow-on-eks/main.tf
```diff
@@ -0,0 +1,542 @@
...
+  name                    = "postgres"
+  recovery_window_in_days = 0 # Set to zero for this example to force delete during Terraform destroy
+}
+
+resource "aws_secretsmanager_secret_version" "postgres" {
+  secret_id     = aws_secretsmanager_secret.postgres.id
+  secret_string = random_password.postgress.result
+}
+
+data "aws_secretsmanager_secret_version" "postgres" {
```
bryantbiggs, line 325: we should be able to safely remove this and use the direct output from the resource itself. anywhere you see `data.aws_secretsmanager_secret_version.postgres.*` can simply remove the `data.` and use `aws_secretsmanager_secret_version.postgres.*`

examples/analytics/airflow-on-eks/main.tf
```diff
@@ -0,0 +1,542 @@
...
+  name                    = "airflow_webserver"
+  recovery_window_in_days = 0 # Set to zero for this example to force delete during Terraform destroy
+}
+
+resource "aws_secretsmanager_secret_version" "airflow_webserver" {
+  secret_id     = aws_secretsmanager_secret.airflow_webserver.id
+  secret_string = random_password.airflow_webserver.result
+}
+
+data "aws_secretsmanager_secret_version" "airflow_webserver" {
```
bryantbiggs, line 350: likewise - we can use the resource directly since its also created here

examples/analytics/airflow-on-eks/main.tf
```diff
@@ -0,0 +1,542 @@
...
+
+#---------------------------------------------------------------
+# Webserver Secret Key
+#---------------------------------------------------------------
+resource "kubectl_manifest" "airflow_webserver" {
+  sensitive_fields = [
+    "data.webserver-secret-key"
+  ]
+
+  yaml_body = <<YAML
```
bryantbiggs, line 364: if you change 
```suggestion
  yaml_body = <<-YAML
```

you can then indent your YAML body and not have leading whitespace in the resulting output (FYI)

examples/analytics/airflow-on-eks/main.tf
```diff
@@ -0,0 +1,542 @@
...
+kind: Secret
+metadata:
+   name: ${local.airflow_webserver_secret_name}
+   namespace: ${module.airflow_irsa.namespace}
+type: Opaque
+data:
+  webserver-secret-key: ${base64encode(data.aws_secretsmanager_secret_version.airflow_webserver.secret_string)}
+YAML
+
+  depends_on = [module.eks_blueprints.eks_cluster_id]
```
bryantbiggs, line 375: because this relies on `module.airflow_irsa.namespace` which relies on the cluster ID, OIDC details, etc., you should be safe to remove the depends here and things will be ordered correctly

examples/analytics/airflow-on-eks/main.tf
```diff
@@ -0,0 +1,542 @@
...
+provisioner: efs.csi.aws.com
+parameters:
+  provisioningMode: efs-ap
+  fileSystemId: ${aws_efs_file_system.efs.id}
+  directoryPerms: "700"
+  gidRangeStart: "1000"
+  gidRangeEnd: "2000"
+YAML
+
+  depends_on = [module.eks_blueprints.eks_cluster_id]
```
bryantbiggs, line 396: we should be able to remove the dependency on the cluster since the `kubectl` provider will wait for it to be ready to receive requests (it polls), but do we need to ensure the EFS CSI driver is installed first? i.e. - do we need to do depend on the addons instead of the cluster itself?

examples/analytics/airflow-on-eks/main.tf
```diff
@@ -0,0 +1,542 @@
...
+  subnet_id       = module.vpc.private_subnets[count.index]
+  security_groups = [aws_security_group.efs.id]
+}
+
+resource "aws_security_group" "efs" {
+  name        = "${local.name}-efs"
+  description = "Allow inbound NFS traffic from private subnets of the VPC"
+  vpc_id      = module.vpc.vpc_id
+
+  ingress {
```
bryantbiggs, line 421: I believe we also should add UDP on 2049 as well

examples/analytics/airflow-on-eks/values.yaml
```diff
@@ -0,0 +1,264 @@
...
+# Airflow Worker Config
+#----------------------------------------------------
+workers:
+  serviceAccount:
+    create: false
+    name: ${airflow_service_account}
+  persistence:
+    enabled: false
+  resources:
+    limits:
```
bryantbiggs, line 63: do we want to provide default requests/limits for workers?

examples/analytics/airflow-on-eks/variables.tf
```diff
@@ -0,0 +1,23 @@
+variable "name" {
```
bryantbiggs, line 1: nit: consistent ordering:

```hcl
variable "foo" {
  description = "Something"
  type        = string
  default     = null
}
```

modules/irsa/outputs.tf
```diff
@@ -1,9 +1,19 @@
...
 
 output "irsa_iam_role_name" {
   description = "IAM role name for your service account"
-  value       = var.irsa_iam_policies != null ? aws_iam_role.irsa[0].name : null
+  value       = try(aws_iam_role.irsa[0].name, null)
+}
+
+output "namespace" {
+  description = "IRSA Namespace"
+  value       = try(kubernetes_namespace_v1.irsa[0].id, var.kubernetes_namespace)
```
bryantbiggs, line 13:
```suggestion
  value       = try(kubernetes_namespace_v1.irsa.metadata[0].name, var.kubernetes_namespace)
```

modules/irsa/outputs.tf
```diff
@@ -1,9 +1,19 @@
...
+}
+
+output "service_account" {
+  description = "IRSA Service Account"
+  value       = try(kubernetes_service_account_v1.irsa[0].id, var.kubernetes_service_account)
```
bryantbiggs, line 18:
```suggestion
  value       = try(kubernetes_service_account_v1.irsa[0].metadata[0].name, var.kubernetes_service_account)
```

bryantbiggs:
overall looks great - my one question would be - from a user/customer perspective, can we do something to better clarify this example vs https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/examples/managed-workflow-apache-airflow 

It doesn't have to be in this PR, but one is an example using a managed service, the other is self-hosting - both great examples but it could be better organized (at least have them in close proximity to one another) and maybe update the name to make it clear that one is showing using the integration with MWAA, the other is self-hosting Airflow but still using integrations into S3/RDS/etc.

examples/analytics/airflow-on-eks/main.tf
vara-bonthu: Yes, I have just added metrics server and cluster autoscaler

examples/analytics/airflow-on-eks/main.tf
vara-bonthu: Nice! Let me add the s3 module

examples/analytics/airflow-on-eks/main.tf
vara-bonthu: done

examples/analytics/airflow-on-eks/main.tf
vara-bonthu: done

examples/analytics/airflow-on-eks/main.tf
vara-bonthu: done

examples/analytics/airflow-on-eks/main.tf
vara-bonthu: Doc is pointing to only TCP. https://docs.aws.amazon.com/efs/latest/ug/accessing-fs-create-security-groups.html
Could you point me to more details here on why UDP is required?

vara-bonthu:
> Agree. We should we move the MWAA example under Analytics folder and then rename the folders to `managed-airflow` and `self-managed-airflow`. 

The main difference is that MWAA has nothing to do with EKS unlike self-managed airflow hosted on EKS. MWAA example is only to trigger the job on EKS

I have added additional information to Docs(apache-airflow.md) and Readme as well.

vara-bonthu:
Check Markdown links / markdown-link-check error is expected as the link is not available in MAIN until this PR is merged

examples/analytics/airflow-on-eks/main.tf
bryantbiggs: Oh, guess I am showing how dated I am 😅 - `While earlier versions of the protocol used User Datagram Protocol (UDP), from version 4.2 onwards NFS runs on top of Transmission Control Protocol (TCP).`

You used to need UDP https://aws.amazon.com/blogs/aws/amazon-efs-update-on-premises-access-via-direct-connect-vpc/ but I guess not anymore, please disregard!

bryantbiggs (approved):
Looks great - few minor details to correct on the CI checks and then good to go !

vara-bonthu:
> Looks great - few minor details to correct on the CI checks and then good to go !

Plan error seems to be related to RAY example. Let me see if that needs fixing

```(and 97 more similar warnings elsewhere)

Error: Error in function call

  on main.tf line 156, in module "eks_blueprints_kubernetes_addons":
 156:     values = [templatefile("${path.module}/helm-values/kube-stack-prometheus-values.yaml", {
 157:       hostname = var.eks_cluster_domain
 158:     })]
    ├────────────────
    │ path.module is "."
    │ var.eks_cluster_domain is null

Call to function "templatefile" failed:
./helm-values/kube-stack-prometheus-values.yaml:6,[27](https://github.com/aws-ia/terraform-aws-eks-blueprints/runs/7971539678?check_suite_focus=true#step:8:28)-[35](https://github.com/aws-ia/terraform-aws-eks-blueprints/runs/7971539678?check_suite_focus=true#step:8:36): Invalid template
interpolation value; The expression result is null. Cannot include a null
value in a string template..
```

PR #882, opened 2022-08-23
feat: Add Datadog operator to Kubernetes addons 

bobdoah:
### What does this PR do?

Adds the [Datadog operator](https://github.com/DataDog/datadog-operator) Helm chart to the Kubernetes addons. This operator can be used to deploy a Datadog agent following recommended best practices to a Kubernetes cluster. 

### Motivation

My company are in the process of moving our monitoring to Datadog. The operator would be a useful addon for us to deploy to our clusters.

### More

### For Moderators


examples/datadog/README.md
```diff
@@ -0,0 +1,82 @@
+# External Secrets Operator Kubernetes addon
```
bryantbiggs, line 1:
```suggestion
# Datadog Operator Kubernetes addon
```

examples/datadog/main.tf
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  tags = local.tags
+}
+
+#---------------------------------------------------------------
+# Datadog agent
+#---------------------------------------------------------------
+
+resource "kubernetes_secret_v1" "datadog_api_key" {
```
bryantbiggs, line 167: the secret and the DatadogAgent manifest should be moved to the addon itself

examples/datadog/main.tf
```diff
@@ -0,0 +1,197 @@
...
+  }
+  data = {
+    # This will reveal a secret in the Terraform state
+    api-key = var.datadog_api_key
+  }
+}
+
+resource "kubectl_manifest" "datadog_agent" {
+  yaml_body  = <<YAML
+apiVersion: datadoghq.com/v1alpha1
```
bryantbiggs, line 180: convert this to a map so that we can do something like:

```hcl
locals {
	datadog_agent = merge(
		  {
	  		apiVersion = "datadoghq.com/v1alpha1"
			kind = "DatadogAgent"
			metadata:
				name = "datadog"
				namespace = ${local.namespace}
			spec =
				clusterName = ${module.eks_blueprints.eks_cluster_id}
				site = "datadoghq.com"
			....
		  },
		var.helm_config.datadog_agent
	)
}
```

Where users can pass in something like

```hcl
	helm_config = {
		datadog_agent = {
			spec = {
				site = "datadoghq.eu"
			}
		}
	}
```

modules/kubernetes-addons/datadog-operator/main.tf
```diff
@@ -0,0 +1,20 @@
...
+  source            = "../helm-addon"
+  manage_via_gitops = var.manage_via_gitops
+  helm_config       = local.helm_config
+  irsa_config       = null
+  addon_context     = var.addon_context
+
+  depends_on = [kubernetes_namespace_v1.this]
+}
+
+resource "kubernetes_namespace_v1" "this" {
```
bryantbiggs, line 11: we can remove this and instead use the `create_namespace` argument provided by the `helm_release` https://registry.terraform.io/providers/hashicorp/helm/latest/docs/resources/release#create_namespace

bryantbiggs:
we'll also need to add docs to the `docs/` directory

examples/datadog/main.tf
bryantbiggs: actually, if we move this to a yaml file, we should be able to read it in as a map and then merge it. we should only add the bare minimum so that users are able to extend to suit

```hcl
locals = {
	datadog_agent = merge(yamldecode(templatefile("${path.module}/datadog_agent.yaml", {
			  cluster_name = module.eks_blueprints.eks_cluster_id
			  namespace   = local.namespace
		 })),
		var.helm_config.datadog_agent
	)
}
```

examples/datadog/main.tf
bobdoah: My only concern with that, is having this _secret_ API key in the Terraform state. I know for a fact it would be undesirable at my company.  

If I were to do this, I'd store the secret in secrets manager or parameter store, and sync it to the cluster.

bobdoah: This is the reason why I included the agent configuration as an example and not in the addon itself.

bryantbiggs: completely agree - but we can provide users the ability to create a secret, or reference a secret. either way a secret has to exist, but we can provide options. the whole premise is that an addon *should be installed with a single flag, but here we are saying users have to install and then semi-manually go and do some additional work

bobdoah: I was viewing the addon as "the operator", but having it handle the agent creation would definitely be preferable. I mean tbh, I wasn't totally sure about dragging the not-really-official kubectl provider into the main code base.

Certainly an optional secret is possible and what I was considering.

bryantbiggs: lets see what the others think - @askulkarni2 / @vara-bonthu / @Zvikan

examples/datadog/README.md
bobdoah: Wouldn't that a be a bit confused?

examples/datadog/main.tf
bobdoah: I refactored this, as suggested, you may have seen. When I tried to pass in a spec element, as above it overwrites the entire spec so elements such as the `clusterName` and `credentials` are lost if they are not explicitly by the user. Terraform doesn't support deep merging of maps like this. So the `spec` key is simply replaced.

bobdoah: @bryantbiggs @askulkarni2 @vara-bonthu @Zvikan  any further thoughts on this refactor and inclusion of the Datadog agent creation?

danderton:
This would be great to get in, is there anything needed to complete this MR?

examples/datadog/main.tf
bryantbiggs: apologies for the delay @bobdoah - after considering both avenues, I think you were correct with your original approach. This is the current practice for CRDs so we will continue with this approach

bryantbiggs (approved):
Thank you @bobdoah and apologies for the long delay. Please feel free to open a PR on https://github.com/aws-samples/eks-blueprints-add-ons to add GitOps support when you get a chance

bobdoah:
Amazing, thanks for fixing this up @bryantbiggs.

bobdoah:
> Please feel free to open a PR on https://github.com/aws-samples/eks-blueprints-add-ons to add GitOps support when you get a chance

Done: https://github.com/aws-samples/eks-blueprints-add-ons/pull/100

PR #883, opened 2022-08-23
Karpenter CRD update

FernandoMiguel:
### What does this PR do?

karpenter requires manual CRD updates https://karpenter.sh/preview/upgrade-guide/#custom-resource-definition-crd-upgrades

### Motivation

v0.15 includes a new CRD resource, `consolidation`

### More

### For Moderators

### Additional Notes

gist output https://gist.github.com/FernandoMiguel/a2c298b1b55d5c3ddf2c3455b7ce3a2d/edit


modules/kubernetes-addons/karpenter/main.tf
```diff
@@ -12,3 +12,25 @@ resource "aws_iam_policy" "karpenter" {
   description = "IAM Policy for Karpenter"
   policy      = data.aws_iam_policy_document.karpenter.json
 }
+
+## karpenter requires manual CRD updates https://karpenter.sh/preview/upgrade-guide/#custom-resource-definition-crd-upgrades
```
bryantbiggs, line 16: I think these need to stay outside the module since they are one-off for users of karpenter <0.14 and they are documented as a manual step. these same actions can be performed by users in their Terraform if they so choose

bryantbiggs (changes requested):

modules/kubernetes-addons/karpenter/main.tf
andrewhibbert: It's not 1.14 specific, it is everytime the CRDs are changed that they need reapplying

FernandoMiguel: As practitioners of IaC having to run steps outside of terraform when some are partially managed with terraform seems counter-intuitive.
If a team has a workflow and is upgrading karpenter per cluster, they would want these changes to CRD to go in so that their Provisioners can benefit from the new features.

If you still feel this is not the best place for this change, I can move it to the example block, but feels out of place to me.

FernandoMiguel: Karpenter team is following a sem.ver release and avoiding breaking changes for the same api version.
That means, to my understanding, that newer versions can be installed and older versions can be downgraded without issues to the provisioner.

bryantbiggs: https://karpenter.sh/preview/upgrade-guide/#how-do-we-break-incompatibility

andrewhibbert: It is a useful feature to be able to apply the CRDs within terraform. I think when you need to apply them after each upgrade it may as well be in terraform and therefore in this module. It can always be enabled via count (set to true), people can disable it if it really is a problem.

andrewhibbert: It is possible for other helm charts too. After install the CRDs aren't managed - https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

FernandoMiguel: @bwagner5 @tzneal Although I understand you might not be particularly comfortable with terraform, your input here would be appreciated.
Following https://karpenter.sh/preview/upgrade-guide/#custom-resource-definition-crd-upgrades and linked to https://helm.sh/docs/chart_best_practices/custom_resource_definitions/
Would you see the replacing of karpenter CRDs for every new version of karpenter as the right approach?
Do you feel it would handle downgrades too ?

thanks in advance

bwagner5: Replacing the Karpenter CRDs has been pretty safe between versions since we haven't had many backwards incompatible changes in v1alpha5. Rollbacks should work fine if the Provisioner spec is also rolled back within the same terrform plan so that the schema matches the version of Karpenter. 

When we do a backwards incompatible change to our CRDs, we'll increment the CRD version and likely provide a conversion webhook. Upgrading the spec to v1alpha6 would mostly be left up to the user. I'm not super familiar with conversion webhooks though, so I may be overlooking something. We also haven't discussed in-depth how we're going to handle the CRD version upgrades.

FernandoMiguel: @bryantbiggs i've bumped the version once again.
I've also added a `depends_on` due to timing reasons where the CRD takes a few moments and that could make karpenter provisioner fail.
Let me know if blueprints is designed to support this pattern of modifying the CRDs when new versions are available, or if best have it in the examples block.
Cheers

bryantbiggs: Thank you - the team has an offsite next week and we have already slated a discussion around CRDs and lifecycle management.

FernandoMiguel: perfect. enjoy the face to face time.

bryantbiggs: Thank you for your patience - after considering how best to approach CRD updates, at this time we do not have a solution that we think would work well for all users in a repeatable manner. So with that, we won't be proceeding with this PR and we'll continue to investigate this area (managing CRDs across version changes and even breaking changes) in search of a solution if one is applicable for EKS Blueprints

FernandoMiguel: Understood.
I'll keep this local.

Is it worth to add it to the examples?

bryantbiggs: Yes, I think that would be a safe compromise with a brief note/reasoning

bryantbiggs:
closing for now

PR #1031, opened 2022-10-06
fix: Let ingress_nginx depend on aws_load_balancer_controller (fixes #1030)

### LINKED ISSUE #1030: Nginx-Ingress should depend on AWS Loadbalancer
## Description

When setting `enable_aws_load_balancer_controller = true` and `enable_aws_load_balancer_controller = true` it is necessary to use targeted destroys to make sure the ingress is being teared down before the loadbalancer as otherwise there will be finalizers left over that cannot be resolved anymore. (see: https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/examples/ingress-controllers/nginx)

This could be resolved by letting the ingress-nginx module explicitly depend on the load balancer module.

## Versions

- Module version: 4.12.1

- Terraform version: 1.2.9

## Reproduction Code

See: https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/examples/ingress-controllers/nginx

## Expected behaviour

`terraform destroy` just works

## Actual behaviour

`terraform destroy` needs to be run multiple times with defined targets

torvitas:
### What does this PR do?

This change makes sure that if you enable the aws_load_balancer_controller and the nginx_ingress the ingress is deleted before the loadbalancer so the finalizers can be fulfilled.

### Motivation

- Resolves #1030 

### More

### For Moderators

### DIFF
--- a/examples/ingress-controllers/nginx/README.md
+++ b/examples/ingress-controllers/nginx/README.md
@@ -88,8 +88,6 @@ The following command destroys the resources created by `terraform apply`
 
 ```shell script
 cd examples/ingress-controllers/nginx
-terraform destroy -target="module.eks_blueprints_kubernetes_addons.module.ingress_nginx[0]" -auto-approve
-terraform destroy -target="module.eks_blueprints_kubernetes_addons.module.aws_load_balancer_controller[0]" -auto-approve
 terraform destroy -target="module.eks-blueprints-kubernetes-addons" -auto-approve
 terraform destroy -target="module.eks-blueprints" -auto-approve
 terraform destroy -auto-approve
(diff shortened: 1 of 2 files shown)


bryantbiggs (changes requested):
We don't want to add a `depends_on` across modules since it results in disruptive operations unnecessarily

torvitas:
Code you maybe elaborate a little more what you mean by disruptive operations?

bryantbiggs:
https://github.com/hashicorp/terraform/issues/26383#issuecomment-699149145

bryantbiggs:
closing for now since proposed solution is not acceptable

Issue #917, opened 2022-08-30
Argo app values is overwritten by addon_config values when add_on_application = true


jochemvankessel:
https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/015834f05273cab73310cec68386aebf7d64627b/modules/kubernetes-addons/argocd/main.tf#L70-L72

See above lines..
I want to pass additional helm values to the  addons chart, like below:

```
    addons = {
      #project            = "cluster-addons"
      path               = "chart"
      repo_url           = "https://github.com/aws-samples/eks-blueprints-add-ons.git"
      add_on_application = true
      values = {
        externalDns = yamldecode(templatefile("${path.module}/addon-values.yaml", {
          hosted_zone     = var.hosted_zone
          assume_role_arn = "arn:aws:iam::xxxxxxxxx:role/dns-manager"
          aws_region      = var.aws_region
          hostname        = var.hosted_zone
          ssl_cert_arn    = data.aws_acm_certificate.issued.arn
        }))["externalDns"]
      }
    }
```
    
Above works as expected when add_on_application = false
However, when add_on_application = true it gets overwritten by the addon_config.
Can we somehow loop through the values for each module and merge that with the addon_config?

askulkarni2:
This is a valid use-case for `external-dns` specifically since in this scenario you are passing AWS resources which could be otherwise created by the consumer terraform module.

jochemvankessel:
Yeah, I think it is valid in all cases you need to pass environmental resource data from the parent to child projects. For sure in one-click deployment scenario’s.
I really do understand the gitops bridge concept, but we could a bit more flexible here I think

askulkarni2:
@jochemvankessel that's great feedback.

cc @kcaws

allamand:
@jochemvankessel does the PR here can be used to solve your issue ?

bryantbiggs:
closed in #1035

runningman:
The merging of helm_config in to the gitops config should be default for all modules. In the current setup there doesn't seem to be a way to override any of the default values from terraform.

PR #1035, opened 2022-10-07
feat: External dns argo values

allamand:
### What does this PR do?

This PR is a response to #917 
It allows additional configuration from Terraform of ExternalDNS add-on deploy with ArgoCD.

### Motivation

When deploying external-dns addon with ArgoCD, we can't provide a custom value file for configuration like we do when deploying with Terraform.

As a workarround, i needed to passe additional parameters to the addons, so this simple fixes allow me do to so, with this step for configuration:

```
  enable_external_dns = true

  external_dns_helm_config = {
    txtOwnerId   = local.name
    zoneIdFilter = data.aws_route53_zone.sub.zone_id # Note: this uses GitOpsBridge
    policy       = "sync"

    #I'm not able to configured additional level values -> Error: Inconsistent conditional result types
    # aws = {
    #   zoneType        = "public"
    #   zonesCacheDuration = "1h"
    # }

    logLevel = "debug"
  }
```

> Note1: It is able to configure the external-dns addon as needed, but i didn't find a solution to configure what is in comment above, this would allow to really configure any value as needed for any add-on

> Note2: This PR needs associated PR in workload to be merged : https://github.com/aws-samples/eks-blueprints-workloads/pull/22

Finally I Have a new example for the moment running in separate repo, but would like to merged it in `examples/upgrade` directory here : https://github.com/allamand/eks-blueprint-terraform-demo/tree/blue-green

- [x] I have Fully tested the eks-cluster-with-external-dns
- [x] I have fully tested the gitops/argocd example (ongoing)

### More

### For Moderators


examples/eks-cluster-with-external-dns/main.tf
```diff
@@ -25,15 +25,31 @@ data "aws_acm_certificate" "issued" {
...
+  name = var.eks_cluster_domain
+}
+
 data "aws_availability_zones" "available" {}
 
 locals {
-  name   = basename(path.cwd)
-  region = "us-west-2"
+  name   = "dns"
+  region = "eu-west-1"
```
askulkarni2, line 36:
```suggestion
  name   = "external-dns"
  region = "us-west-2"
```

examples/eks-cluster-with-external-dns/main.tf
```diff
@@ -25,15 +25,31 @@ data "aws_acm_certificate" "issued" {
...
 
   vpc_cidr = "10.0.0.0/16"
   azs      = slice(data.aws_availability_zones.available.names, 0, 3)
 
+  env                    = "dev"
+
+  # Can be changed to default aws-samples configuration once PR https://github.com/aws-samples/eks-blueprints-workloads/pull/22 is merged
+  workload_repo_url      = "https://github.com/seb-tmp/eks-blueprints-workloads.git"
```
askulkarni2, line 44: Reminder to update :)

askulkarni2 (commented):
@allamand looking good. We likely need an override [here](https://github.com/aws-samples/eks-blueprints-add-ons/blob/eba0ef8e4a7edf761d5428ce1b264bf90e8517f8/chart/templates/external-dns.yaml#L19) as well no?

allamand:
For the parameters I tested I didn't need additional overwrite. Thanks is there a parameter you have in mind ?

modules/kubernetes-addons/external-dns/locals.tf
```diff
@@ -43,8 +43,13 @@ locals {
     irsa_iam_policies                 = concat([aws_iam_policy.external_dns.arn], var.irsa_policies)
   }
 
-  argocd_gitops_config = {
+  default_argocd_gitops_config = {
     enable             = true
     serviceAccountName = local.service_account_name
   }
+
+  argocd_gitops_config = merge(
```
bryantbiggs, line 51: we've never done this and I think it will get quite messy with splatting out all of the values passed via `helm_config`. I believe the original intent of the bridge between Terraform and GitOps was intended to provide only the information that is created in Terraform (i.e. - IAM roles, ACM certs, etc.) to the GitOps operator. It was not designed to specify values in Terraform which are passed to the GitOps operator - those should be defined in the respective manifest repo alongside the templates/manifests that consume them

@askulkarni2 is that accurate or no?

allamand: That may still the case with this evolution, except that we can pass dynamic values, without having to hard coded them in both git repository.

This is working, and i tested it in many different way, it's not perfect solution (it doens't allowed you to pass real values.yaml, but only root values), but it has the advantage of being dynamic (no need to synchronised both repos)

allamand: @askulkarni2 @bryantbiggs can we validate thoses changes ?

allamand: @askulkarni2 @bryantbiggs can we statuate on this ?
I Agrre that in the first place we can configure this values in the addon git repo, but I think it is a nice way to be able to specify some values directly from the Terraform code. This can simplify integration for customers. Please advise on this ? I also need to do the same thing for other controllers like the Argo rollout. In the Workshop there is code to activate the dashboard loadbalancer in terraform, but this didn't work because it is never passed to Argo helm chart.

askulkarni2: I think as far as AWS resource values are concerned (ARNs, Hosted Zone ids, OIDC providers, etc.) this is a valid use case. This is based on the assumption that platform teams usually create the cloud infrastructure required by the applications via IaC. GitOps bridge was designed for this. Unless we can come up with a better way to hand off the values from IaC state to Git state we can move forward with this approach to unblock customers. Longer term, we probably need a better approach where we can externalize the state for such dynamic resource ids and yet not break from gitops philosophy.

askulkarni2 (approved):
LGTM but example needs an update to fix the workload repo location.

andrewhibbert:
@bryantbiggs This looks like a change that should be present for every addon, i.e. we should be able to provide extra values to argocd, other than ones specified in `argocd_gitops_config`

```
  argocd_gitops_config = merge(
    {
      enable             = true
      serviceAccountName = local.service_account
    },
    var.helm_config
  )
}
```

bryantbiggs:
This is not a change we want for every addon. we made an exception for one or two addons, but in general it really obfuscates where values are originating from and also introduces the issue of values potentially clashing. we are evaluating alternative means of better integration (or passing of data) between Terraform and GitOps operators in the new addons project https://github.com/aws-ia/terraform-aws-eks-blueprints-addons

andrewhibbert:
Ideally you should be able to pass the same values to the terraform driven helm and argocd for each addon. At the moment it would be difficult to use argocd, unless the only thing you want to change falls in line with what is hardcoded in `argocd_gitops_config` for each add on

PR #1078, opened 2022-10-24
feat: Allow users to specify the Windows Server version; `2019` (default) or `2022`

IgorEulalio:
### What does this PR do?
This PR allows clients to override the version of windows server and set the default version to 2022 for clusters in 1.22 or above.

### Motivation
Windows Server 2022 is the new and recommend operation system to run Windows containers on Kubernetes. It accommodates all the new features previously only available on Windows Server SAC versions. The AWS Windows container SME team is already helping and spreading the message for customers to start migrating from 2019 to 2022.
With this new addition to EKS Blueprints, customers can leverage the new Windows Server 2022 on EKS clusters version 1.22+ and maintain their Windows Server 2019 on older EKS cluster versions.
- Resolves #<issue-number>
### More
**I used the example that already exists for windows cluster.**
### For Moderators
### Additional Notes

This is a pontual change to make the default windows server used today more flexible and also push users to use the new version of windows server.


marciogmorales:
Marcio Morales - AWS Windows Container SME, reviewed and worked together with Igor in the proposed solution.

modules/aws-eks-self-managed-node-groups/locals.tf
```diff
@@ -49,16 +51,20 @@ locals {
     var.self_managed_ng
   )
 
+  # WINDOWS CONFIGURATION
+  compatible_with_windows_server_2022 = var.context.cluster_version >= "1.22" ? true : false
```
bryantbiggs, line 55: 1. Can users create clusters with 2019 AMI on 1.22/1.23?
2. Can users create clusters with 2022 AMI on 1.19-1.22?

Also, this is not a valid comparison `var.context.cluster_version >= "1.22"`, this is not a numerical comparison

I think what we want to do is add a variable for the `windows_server_version` and have it default to `2019` but users have the ability to pass in `2022` and then we can get rid of the other changes

IgorEulalio: We are looking to change the default to 2022 instead of 2019, but do not exist AMI for Windows Server 2022 that are compatible with cluster below 1.22, so the default must be 2022 for all clusters that are in 1.22 or above and 2019 for cluster that are below.

modules/aws-eks-self-managed-node-groups/locals.tf
```diff
@@ -49,16 +51,20 @@ locals {
...
+  # WINDOWS CONFIGURATION
+  compatible_with_windows_server_2022 = var.context.cluster_version >= "1.22" ? true : false 
   enable_windows_support = local.self_managed_node_group["launch_template_os"] == "windows"
+  windows_server_version = local.self_managed_node_group["windows_server_version"]
 
   predefined_ami_names = {
     amazonlinux2eks = "amazon-eks-node-${var.context.cluster_version}-*"
     bottlerocket    = "bottlerocket-aws-k8s-${var.context.cluster_version}-x86_64-*"
-    windows         = "Windows_Server-2019-English-Core-EKS_Optimized-${var.context.cluster_version}-*"
+    windows         = "Windows_Server-${local.windows_server_version}-English-Core-EKS_Optimized-${var.context.cluster_version}-*"
```
bryantbiggs, line 62: this is all that we need to change (plus add the variable to support):

```suggestion
    windows         = "Windows_Server-${var.windows_server_version}-English-Core-EKS_Optimized-${var.context.cluster_version}-*"
```

IgorEulalio: That variable doesn't exist **windows_server_version**, it only exists as a local. Since the **self_managed_ng** object only pass all the variables inside it, we are merging this variable that came from this object with the **default_self_managed_ng** from locals.

Clients can override these variable passing that to self_managed_ng object, that will be merged:
```
  self_managed_node_group = merge(
    local.default_self_managed_ng,
    var.self_managed_ng
  )
```

bryantbiggs: Correct, thats why I stated `plus add the variable to support` - we need to add this variable to allow users to specify which version they want to use - `2019` or `2022`, and `2019` should be the default since that is the current behavior

IgorEulalio: This variable should be passed inside the **self_managed_ng** object:
```
variable "self_managed_ng" {
  description = "Map of maps of `eks_self_managed_node_groups` to create"
  type        = any
  default     = {}
}
```

Since this object only forwards all the properties inside of that, i made the change in a way that the client can specify, but if doesn't, the default will be adopted, based on the cluster version (2022 for 1.22 and above or 2019 for below 1.22).

IgorEulalio: That works in the same way that **node_group_name** works for example, you can pass and override it, but if you don't, it will adopt a default value from locals:
```
node_group_name = "m4_on_demand"
```

modules/aws-eks-self-managed-node-groups/locals.tf
IgorEulalio: Is it possible to describe that on the new version disclaimer and make this new version default? This behavior is intentional.

PR #1086, opened 2022-10-25
Add Windows kubelet best practice

marciogmorales:
### What does this PR do?

Kubelet behaves differently on Windows, and some parameters need to be in place in order to provide kubelet with an optimal environment.

--System-reserved memory=1.5Gb / --kube-reserved memory=0.5Gi / --eviction-hard memory.available<200Mi,nodefs.available<10%

Windows do not have an out-of-memory process killer as Linux does. Windows always treat all user-mode memory allocations as virtual, and pagefiles are mandatory. Windows nodes do not overcommit memory for processes. The net effect is that Windows won't reach out of memory conditions the same way Linux does and processes page to disk instead of being subject to out-of-memory (OOM) termination. If memory is over-provisioned and all physical memory is exhausted, then paging can slow down performance. This is very common when not setting memory limits on Windows pods, and applications which doesn't recycle memory may starve the entire node memory.

--windows-priorityclass=ABOVE_NORMAL_PRIORITY_CLASS

Windows can limit the amount of CPU time allocated for different processes but cannot guarantee a minimum amount of CPU time. On Windows, the kubelet supports a command-line flag to set the scheduling priority of the kubelet process: --windows-priorityclass. This flag allows the kubelet process to get more CPU time slices when compared to other processes running on the Windows host. More information on the allowable values and their meaning is available at Windows Priority Classes. To ensure that running Pods do not starve the kubelet of CPU cycles, set this flag to ABOVE_NORMAL_PRIORITY_CLASS or above.

### Motivation

Containers administrators usually need to gain deep expertise in Windows Operational Systems to successfully run a heterogeneous Amazon EKS cluster with Windows node group; the motivation is to help them to have Windows best practices already applied when deploying Windows nodes when using EKS Blueprints.

### More

### For Moderators

### Additional Notes

image (4)
image (3)

### DIFF
--- a/modules/launch-templates/templates/userdata-windows.tpl
+++ b/modules/launch-templates/templates/userdata-windows.tpl
@@ -17,7 +17,7 @@ if ($disks_to_adjust -ne $null) {
   }
 }
 
-# Redirect Docker files to new the disk
+# Redirecting Docker files to secondary disk to avoid concurrent I/O during node startup.
 Stop-Service -Name "docker" -Force -NoWait
 
 $dockerdataredirect = @'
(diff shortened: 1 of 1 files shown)


bryantbiggs (changes requested):
lets not hardcode these settings and remove the ability to customize, but this would be a great addition to use as the default settings

marciogmorales:
It makes sense Bryan, I will need to create a condition on Terraform as these values only apply on Windows or we could just add as a comment like [here ](https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/bc1c58ea207d8094a0f95c7e82b80ec038e52ee6/modules/aws-eks-self-managed-node-groups/locals.tf#L14) . The comment approach would be more uncomplicated and easy to understand. What are your thoughts?

bryantbiggs:
yes, I'm not sure exactly how best to add that as a default for windows nodes, but that would be good. or we can just document that its recommended to set those various settings when running windows based nodes

marciogmorales:
Ok,
So I will update the [EKS Blueprint documentation - Windows Self-Managed Node Groups](https://aws-ia.github.io/terraform-aws-eks-blueprints/v4.13.1/node-groups/) with a greater explanation and add on the TF file just a comment like:

_Windows Only: --kube-reserved=memory=0.5Gi --system-reserved=memory=1.5Gi --windows-priorityclass=ABOVE_NORMAL_PRIORITY_CLASS_

Are you ok with this approach?

bryantbiggs:
that sounds great, thank you @marciogmorales

PR #1132, opened 2022-11-04
feat: Update EKS module version and add additional variables supported by the module

bryantbiggs:
### What does this PR do?

- Update EKS module version and add additional variables supported by the module
- Add support for disabling cluster encryption
- Remove workflow scripts that have since moved to a `scripts/` directory

### Motivation

- Internal request

### More

### For Moderators


.github/workflows/iam-policy-generator.py
```diff
@@ -1,35 +0,0 @@
-import json
```
bryantbiggs, line 1: This lives here https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/.github/scripts/iam-policy-generator.py

Zvikan (approved):

.github/workflows/plan-examples.py
```diff
@@ -1,30 +0,0 @@
-import json
```
bryantbiggs, line 1: this lives here https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/.github/scripts/plan-examples.py

main.tf
```diff
@@ -12,77 +18,64 @@ module "kms" {
   tags                    = var.tags
 }
 
-# ---------------------------------------------------------------------------------------------------------------------
-# EKS CONTROL PLANE
-# ---------------------------------------------------------------------------------------------------------------------
 module "aws_eks" {
   source  = "terraform-aws-modules/eks/aws"
-  version = "v18.26.6"
+  version = "v18.29.1"
```
bryantbiggs, line 23: we don't want to go up to v18.30 if we can avoid it - 18.30 added default tags logic that is proving to be problematic and will be removed/reverted in v19.x

askulkarni2 (approved):
🙌🏽

main.tf
```diff
@@ -1,8 +1,14 @@
-# ---------------------------------------------------------------------------------------------------------------------
-# CLUSTER KMS KEY
-# ---------------------------------------------------------------------------------------------------------------------
+locals {
+  cluster_encryption_config = length(var.cluster_encryption_config) > 0 ? [
```
muram, line 2: Hi @bryantbiggs , I believe this conditional should evaluate if the `cluster_encryption_config` variable is not provided which means `== 0` not `> 0`. The default value is an empty list `[]`.

bryantbiggs: ah yes,  you are correct. let me fix this

bryantbiggs: PR https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/1137

Thank you for catching this @muram

PR #1269, opened 2022-12-20
fix: Argocd addons error, bool expressions must have consistent types

botika:
Fixes `The true and false result expressions must have consistent types` 

### More

- [-] Yes, I have added a new example under [examples](https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/examples) to support my PR
- [-] Yes, I have created another PR for add-ons under [add-ons](https://github.com/aws-samples/eks-blueprints-add-ons) repo (if applicable)
- [-] Yes, I have updated the [docs](https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/docs) for this feature

### For Moderators

### DIFF
--- a/modules/kubernetes-addons/argocd/main.tf
+++ b/modules/kubernetes-addons/argocd/main.tf
@@ -69,7 +69,7 @@ resource "helm_release" "argocd_application" {
       { repoUrl = each.value.repo_url },
       each.value.values,
       local.global_application_values,
-      each.value.add_on_application ? var.addon_config : {}
+      each.value.add_on_application ? var.addon_config : null
     ))
     type = "auto"
   }


bryantbiggs:
thank you - please start with an issue and a reproduction + output of the error so we can properly triage

richardj-bsquare:
I get this too, just deploy a blueprint with argocd applications disabled and then set one to true and redeploy...

Here, I turned on external-dns

```
Error: Inconsistent conditional result types
│ 
│   on .terraform/modules/eks_blueprints_kubernetes_addons/modules/kubernetes-addons/argocd/main.tf line 72, in resource "helm_release" "argocd_application":
│   72:       each.value.add_on_application ? var.addon_config : {}
│     ├────────────────
│     │ each.value.add_on_application is false
│     │ var.addon_config is object with 2 attributes
│ 
│ The true and false result expressions must have consistent types. The 'true' value includes object
│ attribute "externalDns", which is absent in the 'false' value.
```

richardj-bsquare:
The workaround is to set the helm_config for the particular application first, and then toggle the enable flag. Changing the configuration and the enabled flag at the same time causes the above error.

botika:
@richardj-bsquare  You can follow the flow that they set and redo the PR , otherwise they don't merge it and I need it but I don't have time. Thanks!

PR #1447, opened 2023-02-22
fix: Use `most_recent = true` when using VPC CNI custom configuration

bryantbiggs:
### What does this PR do?

- Use `most_recent = true` when using VPC CNI custom configuration

### Motivation

- Without using the most recent, the default version of the VPC CNI pulled may or may not have support for the environment variable versions used. This ensures that the latest addon version is used which has access to all of the environment variables used in the examples

### More

### For Moderators


csantanapr (approved):
/lgtm

Issue #1421, opened 2023-02-08
Moving forward with v5 of EKS Blueprints


fcarta29:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

This issue is being created to provide community notice and to help track the work and cutover to v5 of EKS Blueprints. More details of the changes can be found here:
~~[Direction for v5 of EKS Blueprints](https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/DIRECTION_v5.md)~~
[Motivation for v5 of EKS Blueprints](https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/docs/v4-to-v5/motivation.md)

Jumziey:
Do you have any ambitions when it comes to migrating from v4 to v5? It seems it will get complicated given the goals :thinking: 

That said, great initiative! We've been bothered by the same issues you describe and it seems like the correct way forward.

mk2134226:
Any idea how long it will take for v5 to be production ready ? and how long you guys will maintain / patch / update v4 ?

fcarta29:
###FAQ for v5

**When will v5 be GA and ready for production**
_EKS Blueprints are community driven examples of how to build on AWS EKS. Only AWS services like AWS EKS have formal release dates, are officially supported, and are certified as production ready._

**How long you guys will v4 be maintained and updated**
_Once v5 reaches development/refactoring readiness and testing is complete, notice will be made to the community via this [issue](https://github.com/aws-ia/terraform-aws-eks-blueprints/issues/1421) and on the top level [README.md](https://github.com/aws-ia/terraform-aws-eks-blueprints#readme) regarding when cross over from v4 to v5 will be merged. From this point v4 will be tagged/branched and left for those who desire to remain on v4 and for historical context. No further development/updates will occur for v4 and all future changes will only occur on the main v5 version._

Hokwang:
@bryantbiggs Hi, I have one question about v5.
[...]

PR #1486, opened 2023-03-08
fix: Align creation of k8s namespaces and service account in addons

atorrescogollo:
### What does this PR do?

Some addons support setting `helm_config.create_namespace`, but don't support setting `helm_config.create_service_account`. This makes disabling `create_namespace` useless when the creation of the service account is enabled since **the service account can't be created into a non-existent namespace**.

On the other hand, since Blueprints v5 will have different repositories per addon, I would prefer to manage all Kubernetes resources through a custom ArgoCD app. And then, I want **blueprints to only create the IAM roles for IRSA**.

For both reasons, this PR will allow conditionally creating namespaces (already supported) and service accounts per addon.

### Motivation

- Related to #1421

### More

### For Moderators


bryantbiggs:
thank you for the PR - this will be resolved in the new addons module https://github.com/aws-ia/terraform-aws-eks-blueprints-addons

PR #1512, opened 2023-03-27
chore: Use helm_release create_namespace option

kahirokunn:
### What does this PR do?

I have entrusted helm_release with the creation of the ArgoCD namespace.

### Motivation

Kubernetes namespace resources can get stuck when deleted.
For example, in the case of ArgoCD, when deleting an Application resource, if ArgoCD is deleted first, the Application resource Finalizer will stop and the ArgoCD namespace cannot be deleted forever.
Therefore, it is better to leave the creation of the ArgoCD namespace to helm_release, which is especially stable when deleting.

### More

### For Moderators

### DIFF
--- a/modules/kubernetes-addons/argocd/main.tf
+++ b/modules/kubernetes-addons/argocd/main.tf
@@ -7,23 +7,17 @@ module "helm_addon" {
   depends_on = [kubernetes_namespace_v1.this]
 }
 
-resource "kubernetes_namespace_v1" "this" {
-  count = try(local.helm_config["create_namespace"], true) && local.helm_config["namespace"] != "kube-system" ? 1 : 0
-  metadata {
-    name = local.helm_config["namespace"]
-  }
-}
-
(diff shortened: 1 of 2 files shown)


modules/kubernetes-addons/argocd/main.tf
```diff
@@ -7,23 +7,17 @@ module "helm_addon" {
   depends_on = [kubernetes_namespace_v1.this]
 }
 
-resource "kubernetes_namespace_v1" "this" {
```
bryantbiggs, line 10: this would be a breaking change and remove functionality for users today

bryantbiggs:
Thank you for the PR - however, we are in the process of migrating addons to be controlled [here](https://github.com/aws-ia/terraform-aws-eks-blueprints-addons) which if you refer to the new implementation, this will be covered (as seen [here](https://github.com/aws-ia/terraform-aws-eks-blueprints-addons/blob/65f64320390ceb7a3258ca9a1fca0352d65648dc/main.tf#L58))

PR #1518, opened 2023-03-29
fix: Escape regex validation backslash

naxgrp-johnthompson:
### What does this PR do?
This fixes a recent change that broke the argocd addon variables file -- i escaped the backslash before the dot so that it's a valid regex.

### DIFF
--- a/modules/kubernetes-addons/argocd/variables.tf
+++ b/modules/kubernetes-addons/argocd/variables.tf
@@ -10,7 +10,7 @@ variable "applications" {
   default     = {}
 
   validation {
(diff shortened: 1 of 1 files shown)


naxgrp-johnthompson:
terraform first tries to escape the string before it parses the regex, so you need to escape the escape character.

cole-ratner:
this is affecting production for a major aws migration effort for Sovrn as well, please push this through

askulkarni2:
We are looking into it, as a best practice please pin your module usage to the latest released version `4.26.0`.

bryantbiggs (approved):

PR #1540, opened 2023-04-06
fix: Fix spot instance issue in self managed node group template

### LINKED ISSUE #1539: Spot not working for self managed node
## Description

Given there is a node config in Terraform
```
   self_managed_protected = {
    node_group_name    = "self_managed_protected"
    instance_type      = "m4.large"
    subnet_type        = "private"
    capacity_type      = "spot"
    launch_template_os = "amazonlinux2eks"
    capacity_rebalance = false #https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-capacity-rebalancing.html
    desired_size       = 3
    min_size           = 3
    max_size           = 5 # minimum 1
    disk_size          = 30
    pod_pid_limit      = 200
```

Then a node with lifecycle type spot should be created.

The bug is in this file https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/modules/aws-eks-self-managed-node-groups/self-managed-launch-templates.tf

This line needs a fix

```      
capacity_type          = local.self_managed_node_group["capacity_type"] == "spot" ? "" : local.self_managed_node_group["capacity_type"]
```

Instead of the above it should be something like this:
```      
capacity_type          = local.self_managed_node_group["capacity_type"] == "spot" ? "spot" : local.self_managed_node_group["capacity_type"]
```

gauravchaudhary999:
Fixed spot instance bug in self managed node group

- Resolves #1539

### DIFF
--- a/modules/aws-eks-self-managed-node-groups/self-managed-launch-templates.tf
+++ b/modules/aws-eks-self-managed-node-groups/self-managed-launch-templates.tf
@@ -8,7 +8,7 @@ module "launch_template_self_managed_ng" {
       launch_template_os     = local.self_managed_node_group["launch_template_os"]
       launch_template_prefix = local.self_managed_node_group["node_group_name"]
       instance_type          = local.self_managed_node_group["instance_type"]
-      capacity_type          = local.self_managed_node_group["capacity_type"] == "spot" ? "" : local.self_managed_node_group["capacity_type"]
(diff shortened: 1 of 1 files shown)


askulkarni2 (approved):
@gauravchaudhary999 thanks for the PR. LGTM!

Issue #1562, opened 2023-04-25
Create gp3 storage class when enabling ebs csi driver


idjemaoune:
We want to add the ability to create new storage class on the Kubernetes EKS cluster to meet specific storage needs for our applications.

When activating the ebs csi driver, which is necessary to use gp3, we can create a new storage class without making it a default class. So we can keep gp2 and gp3 for workloads that need optimized storage.

bryantbiggs:
see comment here https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/1564#issuecomment-1522328148

PR #1564, opened 2023-04-25
Enable GP3 storage class when enabling ebs csi driver

idjemaoune:
#1562


bryantbiggs:
thank you for the PR! I don't believe we need to embed this into the addon since there are various ways in which the storage class can be created (i.e - Terraform, kubectl, ArgoCD, etc.). Instead, we have demonstrated in the [`stateful`](https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/35d131775b0c2f9235fa880bf171e935b8b50ce3/examples/stateful/main.tf#L216-L259) example how you can add the gp3 storage class and set it as the default storage class while retaining the gp2 storage class

PR #1580, opened 2023-05-04
chore: Validate velero, efs, and stateful example

csantanapr:
Validate velero, efs, and stateful example


bryantbiggs (approved):
fantastic!

PR #1604, opened 2023-05-13
fix(chore): Argo example update

candonov:
## :bangbang: PLEASE READ THIS FIRST :bangbang:

The direction for EKS Blueprints will soon shift from providing an all-encompassing, monolithic "framework" and instead focus more on how users can organize a set of modular components to create the desired solution on Amazon EKS. We have updated the [examples](https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/examples) to show how we use the https://github.com/terraform-aws-modules/terraform-aws-eks for EKS cluster and node group creation. We will not be accepting any PRs that apply to EKS cluster or node group creation process. Any such PR may be closed by the maintainers.

We are hitting also the pause button on new add-on creations at this time until a future roadmap for add-ons is finalized. Please do not submit new add-on PRs. Any such PR may be closed by the maintainers.

Please track progress, learn what's new and how the migration path would look like to upgrade your current Terraform deployments. We welcome the EKS Blueprints community to continue the discussion in issue https://github.com/aws-ia/terraform-aws-eks-blueprints/issues/1421

### What does this PR do?

Reverts ArgoCD to v4

### Motivation

- Resolves #<issue-number>
Reverts ArgoCD example to v4 due to lack of gitops bridge solution in v5. 

### More

### For Moderators

### DIFF
--- a/examples/argocd/README.md
+++ b/examples/argocd/README.md
@@ -45,7 +45,7 @@ The following command will update the `kubeconfig` on your local machine and all
 1. Run `update-kubeconfig` command:
 
     ```sh
-    aws eks --region <REGION> update-kubeconfig --name <CLUSTER_NAME>
+    aws eks --region <REGION> update-kubeconfig --name <CLUSTER_NAME> --alias <CLUSTER_NAME>
     ```
 
 2. List out the pods running currently:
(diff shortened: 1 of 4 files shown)


examples/argocd/README.md
```diff
@@ -45,73 +45,77 @@ The following command will update the `kubeconfig` on your local machine and all
 1. Run `update-kubeconfig` command:
 
     ```sh
-    aws eks --region <REGION> update-kubeconfig --name <CLUSTER_NAME>
+    aws eks --region <REGION> update-kubeconfig --name <CLUSTER_NAME> --alias <CLUSTER_NAME>
```
bryantbiggs, line 48: nit: What is the motivation for the `--alias` flag? Only reason I ask is for consistency - if we are going to use it in the examples, then we should use it across all of the examples as well as the generated output

examples/argocd/main.tf
```diff
@@ -1,6 +1,4 @@
-provider "aws" {
-  region = local.region
-}
+provider "aws" {}
```
bryantbiggs, line 1: lets keep the region so that its known where the resources are deployed. without it, its up to the default fallbacks (i.e. - IAM profile, environment variables, etc.) as to where the resource are created

examples/argocd/main.tf
```diff
@@ -58,6 +55,15 @@ module "eks" {
   cluster_version                = "1.25"
```
bryantbiggs, line 55:
```suggestion
  cluster_version                = "1.26"
```

examples/argocd/main.tf
```diff
@@ -58,6 +55,15 @@ module "eks" {
...
   cluster_endpoint_public_access = true
 
+  # EKS Addons
+  cluster_addons = {
```
bryantbiggs, line 59: doesn't Argo use PVs for Dex/Redis (in HA mode)? We might want to add the EBS CSI driver since thats required now starting in EKS 1.23+

examples/argocd/README.md
candonov: If you don't have the `--alias` it add the cluster to your context with the full ARN and when you are switching contexts it is unnecessary convoluted. `--alias` is not a necessity but definitely recommend to adopt it as a pattern to the rest of the examples.

askulkarni2 (approved):

PR #1607, opened 2023-05-17
fix: Kubecost Container Version health with EKS 1.25+

olegTarassov:
EKS version 1.25+ breaks cost-model container
issue makes pod unhealthy because of error:
```
HorizontalPodAutoscaler v2beta1 is deprecated
```
ref: https://github.com/tektoncd/pipeline/issues/5128

### What does this PR do?

Bumps container tag for cost-model that has the new HPA ApiVersion support with EKS version 1.25+. Without it, container is never in healthy state by kubelet.

### Motivation

related to this: https://github.com/tektoncd/pipeline/issues/5128

### More

- [X ] Yes, I have tested the PR using my local account setup (Provide any test evidence report under Additional Notes)

### For Moderators


askulkarni2 (approved):
Thanks for the PR!

PR #1642, opened 2023-06-08
fix: Correct `teams` patterns use of multi-team namespace creation

sylwit:
The readme says 2 namespaces team-blue and team-red will be created, but the code has some hardcoded values instead of using the key from the iterator


bryantbiggs (approved):
thank you

PR #1648, opened 2023-06-12
feat: Inverse tags merge order to allow overriding hardcoded values

### LINKED ISSUE #1647: Allow the override of hardcoded ASGs tags
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

#### What is the outcome that you are trying to reach?

We have a use case where the Cluster Autoscaler needs to be disabled on a specific Node Group.

The Cluster Autoscaler Autodiscovery relies on Autoscaling Group tags.
The `k8s.io/cluster-autoscaler/enabled` tag is currently hardcoded to "true" in the EKS Blue Prints, for both the Managed Node Groups and the Self-Managed Node Groups:
* https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/v4.32.1/modules/aws-eks-managed-node-groups/locals.tf
* https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/v4.32.1/modules/aws-eks-self-managed-node-groups/locals.tf#L84

Due to the order of the merge, there is currently no way to override hardcoded values to set the `k8s.io/cluster-autoscaler/enabled` tag to false. 

#### Describe the solution you would like

For Self-Managed Node Groups, inverse the tags merge order. This would keep the existing behaviour while allowing overrides of hardcoded values through the `self_managed_node_group["additional_tags"]` attribute.

See PR https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/1648

For Managed Node Groups, tags are not propagated to created ASGs, overriding hardcoded values would not have any effect on the ASGs.

#### Describe alternatives you have considered

* Use the `aws_autoscaling_group_tag` TF resource to tag Managed Node Group ASGs
  * Managed Node Group ASGs IDs are not known before Managed Node Groups deployments. Since tags are dynamic, TF cannot plan until managed node groups are deployed.
* Adding additional tags to Self-Managed groups and configuring CA to auto-discover ASGs based on those additional tags
  * As we can't add additional tags to Managed Node Groups ASGs, this would force to rely on Self-Managed Node Groups exclusively.

#### Additional context

* [autoscaler/README.md at master · kubernetes/autoscaler · GitHub](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md)

AdrienKuhn:
# Description

### Motivation and Context

- Resolves #1647

- [x] ~Yes, I have updated the [docs](https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/docs) for this feature~ N/A

### DIFF
--- a/modules/aws-eks-self-managed-node-groups/locals.tf
+++ b/modules/aws-eks-self-managed-node-groups/locals.tf
@@ -85,5 +85,6 @@ locals {
       "k8s.io/cluster-autoscaler/node-template/label/eks/nodegroup"    = local.self_managed_node_group["node_group_name"]
       "managed-by"                                                     = "terraform-aws-eks-blueprints"
     },
-    local.self_managed_node_group["additional_tags"])
+    local.self_managed_node_group["additional_tags"]
+  )
 }
--- a/modules/aws-eks-self-managed-node-groups/locals.tf
+++ b/modules/aws-eks-self-managed-node-groups/locals.tf
@@ -76,7 +76,6 @@ locals {
 
   common_tags = merge(
     var.context.tags,
-    local.self_managed_node_group["additional_tags"],
     {
       Name                                                             = "${local.self_managed_node_group["node_group_name"]}-${var.context.eks_cluster_id}"
       "kubernetes.io/cluster/${var.context.eks_cluster_id}"            = "owned"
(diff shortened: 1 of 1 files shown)


bryantbiggs:
thank you for the PR - however, we are no longer accepting changes for v4.x and encourage users to to upgrade to use the https://github.com/terraform-aws-modules/terraform-aws-eks module instead

PR #1652, opened 2023-06-16
fix: use local irsa module to create k8s service account resources

AureaMohammadAlavi:
# Description

Unlike the local `../../modules/irsa` module, the `terraform-aws-modules/iam/aws//modules/iam-role-for-service-accounts-eks` module does not create Kubernetes service account resources. Since no service accounts are being created, `ExternalSecret` resources will be created with an `InvalidProviderConfig` status.

### Motivation and Context

- Resolves #<issue-number>

### DIFF
--- a/examples/external-secrets/main.tf
+++ b/examples/external-secrets/main.tf
@@ -119,7 +119,8 @@ module "eks_blueprints_addons" {
   # Add-ons
   enable_external_secrets = true
 
-  tags = local.tags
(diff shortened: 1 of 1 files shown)


bryantbiggs:
Thank you for the PR but we are moving off of the local modules. If you are encountering an issue, please start with a GitHub issue first so we can properly discuss and triage before moving to make any changes

PR #1695, opened 2023-07-21
feat: Adding Istio example for EKS blueprint

praseedasathaye:
# Description
A blueprint example that streamlines the process of deploying Istio within an EKS cluster. Integrating Istio with Amazon Elastic Kubernetes Service (EKS) using EKS Blueprint is essential to unlock the full potential of both technologies and enhance the capabilities of cloud-native applications. 

### Motivation and Context
Istio brings advanced service mesh capabilities by empowering developers to manage and secure microservices with ease. By integrating Istio into the EKS blueprint, organizations can take advantage of its traffic management, security, and observability features seamlessly within their Kubernetes clusters, leading to enhanced control and flexibility. EKS blueprint with Istio simplifies the deployment and management process, reducing the complexity and time required to set up Kubernetes environment with Istio. The motivation to develop an EKS blueprint with Istio setup lies in providing developers/users with a robust, efficient, and easy-to-use platform for deploying and managing cloud-native applications while harnessing the full potential of Istio's service mesh capabilities within the AWS ecosystem. 

### Additional Notes
No


examples/istio/main.tf
```diff
@@ -0,0 +1,217 @@
...
+  }
+}
+
+resource "helm_release" "istio_base" {
+  repository = local.istio_chart_url
+  chart      = "base"
+  name       = "istio-base"
+  namespace  = kubernetes_namespace.istio_system.metadata[0].name
+  version    = local.istio_chart_version
+  wait       = false
```
bryantbiggs, line 144: we have found that letting Kubernetes resolve the deployments in more of an "eventually consistent" manner works better than forcing Terraform to wait for the deployments to succeed.

examples/istio/main.tf
```diff
@@ -0,0 +1,217 @@
...
+  depends_on = [
+    module.eks_blueprints_addons
+  ]
+}
+
+resource "helm_release" "istiod" {
+  repository = local.istio_chart_url
+  chart      = "istiod"
+  name       = "istiod"
+  namespace  = helm_release.istio_base.metadata[0].namespace
```
bryantbiggs, line 155: We can chain the outputs to leave a dependency trail for Terraform to follow and avoid the explicit `depends_on`

examples/istio/main.tf
```diff
@@ -0,0 +1,217 @@
...
+      args = ["eks", "get-token", "--cluster-name", module.eks.cluster_name]
+    }
+  }
+}
+
+data "aws_availability_zones" "available" {}
+
+locals {
+  name   = basename(path.cwd)
+  region = "us-west-2"
```
bryantbiggs, line 35: We only want to provide variables when we require information from users to deploy the pattern (I.e. - domain name or Route53 hosted zone ID). If we provide variables this now exposes the pattern as something that can be used in-place which makes it look and feel like a module. However, we do not provide any semantic version guarantees on patterns and they can change at any time to reflect current practices - not exposing variables avoids exposing that "interface" and makes the distinction clear that these are not modules

bryantbiggs (approved):
Looks awesome, thank you! Left a few comments just FYI

Issue #1630, opened 2023-06-05
Started getting this error subdir "modules/kubernetes-addons/helm-addon" not found


mk2134226:
Everything was ok this morning but now started seeing this error

subdir "modules/kubernetes-addons/helm-addon" not found

awarner-greshamtech:
Same. We pin the module version:
```
module "eks_blueprints" {
  source = "github.com/aws-ia/terraform-aws-eks-blueprints?ref=v4.27.0"
```

vale21:
Same for us. 
We noticed that even if we pint a specific tag (v4.27.0) and sometimes it dowloads the correct version:

Downloading git::https://github.com/aws-ia/terraform-aws-eks-blueprints.git?ref=v4.27.0 for eks-cluster.eks_blueprints_kubernetes_addons...

there are a couple of internal modules that are downloading main:

Downloading git::https://github.com/aws-ia/terraform-aws-eks-blueprints.git for eks-cluster.eks_blueprints_kubernetes_addons.portworx.helm_addon...

Downloading git::https://github.com/aws-ia/terraform-aws-eks-blueprints.git for eks-cluster.eks_blueprints_kubernetes_addons.sysdig_agent.helm_addon...

alexoughton:
I have found that this is due to calling external modules which do not pin to a specific release. For example:

https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/80bd3b92e0f82f45a187805d6d847db320a01c31/modules/kubernetes-addons/main.tf#L391

which calls:

https://github.com/portworx/terraform-eksblueprints-portworx-addon/blob/bf6d1f68eee97052cac105563bebf19a5e6f4512/main.tf#L2

The problem is commit `19c7cfd` to `main` which removes the referenced directory in the external module. It appears that a coordination is required here to fix this in external modules which are called, and perhaps also publish another 4.x release to call those fixes.

DanielCalvo:
Same here -- did an apply just fine this morning EU time and now its broken.

yehoshoua:
Same here we use TAG 4.29

yebolenko:
We have the same issue !!! Please fix. It also happens if we clone it locally (based on the tag).
If we use the local (cloned) repo, we still receive 
 Error: subdir "modules/kubernetes-addons/helm-addon" not found

bryantbiggs:
[...]

PR #1699, opened 2023-07-24
chore: Remove remaining modules from project, update workflows now that modules are removed

bryantbiggs:
# Description

- Remove remaining modules from project
- Update (simplify) workflows now that modules are removed; CI checks should run faster now 🎉 

### Motivation and Context

- Its been almost [2 months since](https://github.com/aws-ia/terraform-aws-eks-blueprints/issues/1630#issuecomment-1577525242) these were re-added temporarily while we ensured partner addons were updated and give users time to migrate onto the v4.32.1 version. With this removal, we can hoist the patterns to the root of the repo and remove checks for modules and instead focus only on patterns in this project
	- Ref https://github.com/aws-ia/terraform-aws-eks-blueprints/issues/1630#issuecomment-1577525242


.github/workflows/pre-commit.yml
```diff
@@ -87,71 +84,10 @@ jobs:
...
-      - name: Pre-commit Terraform ${{ steps.minMax.outputs.minVersion }}
-        uses: clowdhaus/terraform-composite-actions/pre-commit@v1.8.3
-        # Run only validate pre-commit check on min version supported
-        if: ${{ matrix.directory ==  '.' && steps.changes.outputs.src== 'true' }}
-        with:
-          terraform-version: ${{ steps.minMax.outputs.minVersion }}
-          tflint-version: ${{ env.TFLINT_VERSION }}
-          args: 'terraform_validate --color=always --show-diff-on-failure --files $(ls *.tf)'
-
-  preCommitMaxVersion:
```
bryantbiggs, line 107: This was running `pre-commit run -a` from the root of the project. Now we are running this per pattern which means we can remove this and also speed up the CI checks

askulkarni2 (approved):

PR #1728, opened 2023-08-14
fix: Enable EventBridge rules/targets before EKS cluster

vchintal:
# Description

The eventbridge target for the rule `eks-api-endpoint-create` wasn't getting created with the first `terraform apply` command and it doesn't get created until after the EKS cluster is completely setup causing the initial ENI create events to be missed.

This PR ensures that the first  `terraform apply` command:
1.  Includes the target `module.eventbridge` that takes care of creating both modules `module.create_eni_lambda` and `module.delete_eni_lambda`
2. For the above to be possible, it removes the dependency of `module.delete_eni_lambda` on `module.eks`

### DIFF
--- a/examples/privatelink-access/README.md
+++ b/examples/privatelink-access/README.md
@@ -21,7 +21,7 @@ created so that it can respond to the ENIs created by the EKS control plane:
 
 ```sh
 terraform init
-terraform apply -target=module.create_eni_lambda -target=module.nlb
+terraform apply -target=module.eventbridge -target=module.nlb
 ```
 
 Enter `yes` at command prompt to apply
--- a/.github/scripts/plan-examples.py
+++ b/.github/scripts/plan-examples.py
@@ -10,6 +10,7 @@ def get_examples():
     """
     exclude = {
         'examples/appmesh-mtls',  # excluded until Rout53 is setup
+        'examples/privatelink-access',
         'examples/blue-green-upgrade/environment',
         'examples/blue-green-upgrade/modules/eks_cluster'
     }
--- a/examples/privatelink-access/privatelink.tf
+++ b/examples/privatelink-access/privatelink.tf
@@ -218,7 +218,9 @@ module "delete_eni_lambda" {
 
   environment_variables = {
     TARGET_GROUP_ARN = module.nlb.target_group_arns[0]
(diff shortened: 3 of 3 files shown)


askulkarni2 (approved):

askulkarni2 (commented):
@vchintal please see the error in the CI check.

vchintal:
> @vchintal please see the error in the CI check.

Hi @askulkarni2, its a know issue unfortunately. Since we do not run all of the TF at once, we cannot expect the TF plan to succeed. 

If you see in the README we have broken this up into two TF apply commands. If the first command is run and then the TF plan is run, it will succeed. Let me know if this explanation makes sense.

askulkarni2:
Gotcha. Can you please update the github workflow to skip this blueprint then?

vchintal:
@askulkarni2  Sure, can do! Can you please point me to where I have to make the necessary changes? I haven't done this before and do not want to assume.

askulkarni2:
Sure, it would just be under https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/.github/scripts/plan-examples.py.

vchintal:
@askulkarni2  Think I have done it right, please check the latest commit. Thanks 🙏🏼

examples/privatelink-access/README.md
```diff
@@ -21,7 +21,7 @@ created so that it can respond to the ENIs created by the EKS control plane:
 
 ```sh
 terraform init
-terraform apply -target=module.create_eni_lambda -target=module.nlb
```
bryantbiggs, line 24: doesn't the lambda that listens to create cluster events need to be created before the cluster?

bryantbiggs: ah disregard - missed the part about eventbridge requiring lambda so this ensures the lambda is created before the cluster

vchintal: Yes, the `module.eventbridge` has dependency on the both the `module.create_eni_lambda` and `module.delete_eni_lambda`, so it will run both. Not only that once the lambdas are successfully created, the targets of the rules are updated to point to those lambdas. 

What was happening before was, the lambda was in place and so was the rule but the rule target was not pointing to lambda. It was something that the `module.eventbridge` addresses in totality.

PR #1735, opened 2023-08-17
feat: Addition of Istio observability add-ons and enabling cross-zone loadbalancing on Ingress NLB 

vchintal:
### Description
1. Added [jaeger, prometheus, grafana, kiali] observability add-ons
2. Added annotation on the istio-ingress NLB to allow cross-zone loadbalancing


dcyoung:
I tried this PR yesterday, but am encountering some odd behavior.

In the previous code, the "istio-system" namespace  is created by terraform with injection enabled, but after the proposed changes I see that neither `istio-system` nor `istio-ingress` namespaces contain the injection label. Does this matter/is it intentional? 

Additionally, inspecting the target groups in the created load balancer show multiple unhealthy instances. I'm guessing this is because the ingress is deployed to a single node? This was also the case in the original code, and likely not introduced by this PR. However, while changes to this example are being proposed, is there any expected node selection logic or updates to the target groups that would avoid the unhealthy instances?

Lastly, after switching the ingress to a dedicated namespace, incoming traffic is still able to reach the ingress but the ingress is unable to route to the internal destinations defined by destination rules. These rules were working with the previous code, but now result in 503s

```
upstream connect error or disconnect/reset before headers. retried and the latest reset reason: connection failure, transport failure reason: TLS error: 268435581:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED
```

with the istio ingress gateway complaining:
```
[2023-08-18T17:15:55.069Z] "GET / HTTP/1.1" 503 UF,URX upstream_reset_before_response_started{connection_failure,TLS_error:_268435581:SSL_routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED} - "TLS_error:_268435581:SSL_routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED" 0 218 82 - "10.0.45.213" ....
```

I see you have not made changes to the installed samples charts - but are there expected updates to destination rules that i might be missing?

vchintal:
> In the previous code, the "istio-system" namespace is created by terraform with injection enabled, but after the proposed changes I see that neither `istio-system` nor `istio-ingress` namespaces contain the injection label. Does this matter/is it intentional?

It was intentional. The [install documentation](https://istio.io/latest/docs/setup/install/helm/) never actually suggested that `istio-system` have that label. However once `istio` is installed, this label must be there on namespaces where microservices will be deployed that are meant to be part of the mesh.

> Additionally, inspecting the target groups in the created load balancer show multiple unhealthy instances. I'm guessing this is because the ingress is deployed to a single node? This was also the case in the original code, and likely not introduced by this PR. However, while changes to this example are being proposed, is there any expected node selection logic or updates to the target groups that would avoid the unhealthy instances?

I deployed the full `bookinfo` application and tested it out and everything is working just fine. Of all the listeners, the TargetGroup for Protocol:Port of TCP:15021 is all healthy. The TargetGroup for Protocol:Port of TCP:80 was initially all in `unhealthy` state but once the `bookinfo` application is deployed one of the instances started showing up healthy. I am unable to provide further explanation on this as I need to look into this further but it seems the health status is driven applications opening the corresponding node port.

vchintal:
I have further update. The `istio-ingress` LB accepts traffic on three ports and forwards it to the `istio-ingress` pod in `istio-ingress` namespace. Which ports (`443` or `80`) get opened up depends on the type of Gateway one deploys. In my example, [bookinfo-gateway](https://raw.githubusercontent.com/istio/istio/release-1.18/samples/bookinfo/networking/bookinfo-gateway.yaml) opens the traffic on port 80 and hence the instance on which `istio-ingress` deployment was running on ended up being `healthy`. If the `istio-ingress` deployment is scaled to 3 all of the pods land on unique instances and hence all them instantly turned healthy. Hope this explanation helps.

dcyoung:
> Hope this explanation helps.

it does. thanks for the insight!

examples/istio/main.tf
```diff
@@ -2,32 +2,31 @@ provider "aws" {
...
 
-provider "kubernetes" {
+provider "kubectl" {
   host                   = module.eks.cluster_endpoint
   cluster_ca_certificate = base64decode(module.eks.cluster_certificate_authority_data)
+  token                  = data.aws_eks_cluster_auth.this.token
+  load_config_file       = false
+}
 
-  exec {
```
bryantbiggs, line 9: the exec method is the method that is recommended by the Hashicorp providers (Kubernetes, Helm) - lets not change this

examples/istio/main.tf
```diff
@@ -154,21 +157,19 @@ resource "helm_release" "istiod" {
   name       = "istiod"
   namespace  = helm_release.istio_base.metadata[0].namespace
   version    = local.istio_chart_version
-  wait       = false
```
bryantbiggs, line 157: by default, our addons do not wait - instead, we rely on the normal Kubernetes contructs provided by the operators/controllers deployed by the charts to eventually reconcile their state. Lets revert this and remove the explicit depends on for other Istio resources

examples/istio/main.tf
```diff
@@ -178,13 +179,39 @@ resource "helm_release" "istio_ingress" {
...
-            "service.beta.kubernetes.io/aws-load-balancer-scheme" = "internet-facing"
+            "service.beta.kubernetes.io/aws-load-balancer-type"       = "nlb"
+            "service.beta.kubernetes.io/aws-load-balancer-scheme"     = "internet-facing"
+            "service.beta.kubernetes.io/aws-load-balancer-attributes" = "load_balancing.cross_zone.enabled=true"
           }
         }
       }
     )
   ]
+  depends_on = [helm_release.istiod, helm_release.istio_base]
```
bryantbiggs, line 190: lets remove this

examples/istio/main.tf
```diff
@@ -178,13 +179,39 @@ resource "helm_release" "istio_ingress" {
...
+    Accept = "text/plain"
+  }
+}
+
+data "kubectl_file_documents" "istio_addon" {
+  for_each = data.http.istio_addon
+  content  = each.value.response_body
+}
+
+resource "kubectl_manifest" "istio_addon" {
```
bryantbiggs, line 211: we want to minimize, and eventually remove, the use of this provider since has not been updated in some time and looks like it might be abandoned https://github.com/gavinbunney/terraform-provider-kubectl/issues/247

Can we just add instructions in the document to do something like `curl <your file url> | kubectl apply -f -`

such as `curl [<your file url>](https://raw.githubusercontent.com/istio/istio/master/samples/addons/kiali.yaml) | kubectl apply -f -`

vchintal:
Thanks @bryantbiggs the review was clear and extremely helpful.  Though I know I can fix the same PR, I would rather issue a clean one based on the recommendations.

PR #1744, opened 2023-09-04
fix: Add missing AddTags permission for AWS load balancer controller

ghaering:
# Description

Allow load balancer controller to create tags on ELB.

### Motivation and Context

Without the change in this pull request, we could not have load balancers created. According to a colleague, there was a recent change in the AWS APIs related to tagging and load balancers, which is the reason why we need this change.

See for example this for a discussion how it was solved there:
https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3046


askulkarni2:
This has been fixed in https://github.com/aws-ia/terraform-aws-eks-blueprints-addons/releases/tag/v1.6.0.

PR #1770, opened 2023-09-20
feat: Adding a pattern for AWS for FluentBit with ContainerInsights and Kubelet monitoring integration.

rodrigobersa:
# Description

### Motivation and Context

Due the requests of features related to AWS for Fluentbit, and the lack of an existing examples for that, I'm adding this pattern to address those. 


patterns/fluentbit-containerinsights/main.tf
```diff
@@ -0,0 +1,230 @@
...
+    from_port   = 0
+    to_port     = 0
+    protocol    = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  tags = local.tags
+}
+
+resource "aws_vpc_endpoint" "guardduty" {
```
bryantbiggs, line 182: copy+paste error? do we need these guardduty resource?

patterns/fluentbit-containerinsights/README.md
```diff
@@ -0,0 +1,337 @@
+# AWS for FluentBit with ContainerInsights and Kubelet monitoring.
```
bryantbiggs, line 1: I'll offer an alternative thought - what about showing this as a snippet? 

1. Most of the detail is in the README (and its great info), very little has to do with the Terraform
2. I suspect there are a number of different logging configurations/scenarios that users may inquire about
3. There isn't anything specific to the cluster itself

What do you think?

patterns/fluentbit-containerinsights/main.tf
rodrigobersa: No, I wanted to add GuardDuty. We don't really need it, however since we are showing a 
"monitoring" pattern, I think it's interesting to add.

PR #1790, opened 2023-10-11
Missing conf worker data plane logs fluentbit add on

gwmanish:
# Description

### Motivation and Context

- Resolves #<issue-number>

### Additional Notes

 While using fluentbit in EKS, I observed only pods logs are being sent to the cloudwatch. I am suggesting to add configurations for the worker node logs and dataplane logs as well. So I have added the configurations in values.yaml file. Please review and see if its useful.

### DIFF
--- a/.github/workflows/plan-examples.yml
+++ b/.github/workflows/plan-examples.yml
@@ -5,6 +5,7 @@ on:
   pull_request_target:
     branches:
       - main
+      - v4
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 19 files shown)


bryantbiggs:
thank you for the PR! However, we no longer manage modules/addons in this project but instead in the new addons module located at https://github.com/aws-ia/terraform-aws-eks-blueprints-addons

If there is a bug or feature gap, please open an issue there and we can discuss further, thank you!

PR #1799, opened 2023-10-20
Fix: pattern-blue-green-upgrade update example for tfvars

allamand:
# Description

### Motivation and Context

The example tfvars was not using the right source for gitops bridge

- Resolves #<issue-number>


csantanapr:
Thanks for the PR @allamand

patterns/blue-green-upgrade/environment/versions.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">= 1.0"
+  required_version = ">= 1.4.0"
```
bryantbiggs, line 2: are we using any features that require a MSV of 1.4?

bryantbiggs: the required version should be the minimum supported version - so unless we are using a feature thats only available starting with Terraform 1.4, then we default to 1.0

allamand: ok

bryantbiggs (approved):

PR #1803, opened 2023-10-24
feat: Eliminated the use of temporary keys for tests with kubectl in `privatelink-access` pattern

vchintal:
# Description

The `privatelink-access` pattern documentation had steps which involved creating temporary set of AWS credentials for the use of testing PrivateLink connectivity. This approach is manual and copy pasting credentials, however short-lived isn't ideal. 

With the new set of changes introduced in this PR:
1. The EKS cluster would initially come up with its API endpoint as public 
2. This would ensure that the Client EC2 instance's attached role is added as `system:masters` in `aws-auth` configmap
3. Which would allow the Client EC2 to run `kubectl` commands as an `Admin` against the EKS cluster in the Client VPC without the need for any credentials
4. The documentation is also corrected to:
    * Include the right steps to deploy the pattern
    * Include steps to take the cluster API endpoint private after Terraform finishes
    * Include cleaner steps to test the pattern 
    * Add additional step of take the cluster API endpoint public before tearing down the resources


vchintal:
Got additional unplanned files in the commit, closing this for now.

PR #1869, opened 2024-01-30
feat: New Bottlerocket pattern

rodrigobersa:
# Description

### Motivation and Context

As of today there is no patterns of how to run and customize Bottlerocket and it's Update Operator to automate CVE patching.


patterns/bottlerocket/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+This pattern demostrates how to deploy Amazon EKS Clusters using [Bottlerocket OS](https://aws.amazon.com/bottlerocket/) for Managed Node Groups (MNG) and [Karpenter](https://karpenter.sh/), using [Bottlerocket Update Operator (BRUPOP)](https://github.com/bottlerocket-os/bottlerocket-update-operator) to manage CVE patches automatically at the Node OS level. The BROPUP doesn't work with minor or major upgrades, just patch level.
+
+## Deploy
+
+See [here](https://aws-ia.github.io/terraform-aws-eks-blueprints/getting-started/#prerequisites) for the prerequisites and steps to deploy this pattern.
+
+## Validate
+
+* List all Nodes in the cluster. You should see three Nodes that belongs to the defined MNG, and should be in the `v1.27.4-eks-cedffd4` version since we are using a specific AMI ID to teste the BRUPOP.
```
askulkarni2, line 11: Can we use 1.29?

askulkarni2, line 11:
```suggestion
* List all Nodes in the cluster. You should see three Nodes that belongs to the defined MNG, and should be in the `v1.27.4-eks-cedffd4` version since we are using a specific AMI ID to test the BRUPOP.
```

patterns/bottlerocket/addons.tf
```diff
@@ -0,0 +1,152 @@
...
+
+  description   = "CRDs for Bottlerocket Update Operator"
+  chart         = "bottlerocket-shadow"
+  chart_version = "1.0.0"
+  repository    = "https://bottlerocket-os.github.io/bottlerocket-update-operator/"
+
+  depends_on = [module.eks_blueprints_addons]
+}
+
+module "bottlerocket_update_operator" {
```
askulkarni2, line 111: Should we create them in addons?

patterns/bottlerocket/addons.tf
```diff
@@ -0,0 +1,152 @@
+################################################################################
```
askulkarni2, line 1: Why separate `.tf` files?

patterns/bottlerocket/karpenter-resources/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v2
+name: karpenter-resources
+description: Helm chart to deploy Karpenter resources
+version: 0.0.1
```
askulkarni2, line 4: run `pre-commit`

askulkarni2 (commented):
Nice PR! Some comments.

patterns/bottlerocket/README.md
rodrigobersa: Not really, there is just one available AMI for 1.29, so no room for showcasing the upgrade scenario. I can change to 1.28.

patterns/bottlerocket/addons.tf
rodrigobersa: I was thinking the same thing. Give me a couple of days to add the addons there.

patterns/bottlerocket/karpenter-resources/Chart.yaml
rodrigobersa: I missed this dir. Done!

patterns/bottlerocket/README.md
rodrigobersa: Good point! I just updated to v1.28.

patterns/bottlerocket/addons.tf
allamand: Is there a decision on this one ?

patterns/bottlerocket/addons.tf
allamand: Can we also add them in GitOps-bridge ?

patterns/bottlerocket/README.md
```diff
@@ -0,0 +1,151 @@
+# Bottlerocket with Bottlerocket Update Operator
+
+This pattern demostrates how to deploy Amazon EKS Clusters using [Bottlerocket OS](https://aws.amazon.com/bottlerocket/) for Managed Node Groups (MNG) and [Karpenter](https://karpenter.sh/), using [Bottlerocket Update Operator (BRUPOP)](https://github.com/bottlerocket-os/bottlerocket-update-operator) to manage CVE patches automatically at the Node OS level. The BROPUP doesn't work with minor or major upgrades, just patch level.
+
```
allamand, line 4: Gen we add a paragraph on the value of bottlerocket ? Why a customer should care and use bottlerocket for EKS, is it better than AL2 ? Is there some caveats to take into account ?

patterns/bottlerocket/addons.tf
rodrigobersa: Sure. I'll make sure to add the respective variables.

patterns/bottlerocket/addons.tf
rodrigobersa: Not yet.

patterns/bottlerocket/README.md
rodrigobersa: I'm not sure if the blueprints pattern is the place for that. Since we're linking the official docs, customers can use that. But @askulkarni2 or @bryantbiggs can correct me if I'm wrong.

patterns/bottlerocket/addons.tf
rodrigobersa: Done!

rodrigobersa:
Added the `bottlerocket_update_operator` addon.

patterns/bottlerocket/README.md
```diff
@@ -0,0 +1,151 @@
...
+## Deploy
+
+See [here](https://aws-ia.github.io/terraform-aws-eks-blueprints/getting-started/#prerequisites) for the prerequisites and steps to deploy this pattern.
+
+## Validate
+
+* List all Nodes in the cluster. You should see three Nodes that belongs to the defined MNG, and should be in the `v1.28.1-eks-f0272c7` version since we are using a specific AMI ID to test the BRUPOP.
+
+```bash
```
bryantbiggs, line 13: just to be consistent, lets use `sh` like the rest of the patterns - change this throughout

patterns/bottlerocket/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+This pattern demostrates how to deploy Amazon EKS Clusters using [Bottlerocket OS](https://aws.amazon.com/bottlerocket/) for Managed Node Groups (MNG) and [Karpenter](https://karpenter.sh/), using [Bottlerocket Update Operator (BRUPOP)](https://github.com/bottlerocket-os/bottlerocket-update-operator) to manage CVE patches automatically at the Node OS level. The BROPUP doesn't work with minor or major upgrades, just patch level.
+
+## Deploy
+
+See [here](https://aws-ia.github.io/terraform-aws-eks-blueprints/getting-started/#prerequisites) for the prerequisites and steps to deploy this pattern.
+
+## Validate
+
+* List all Nodes in the cluster. You should see three Nodes that belongs to the defined MNG, and should be in the `v1.28.1-eks-f0272c7` version since we are using a specific AMI ID to test the BRUPOP.
```
bryantbiggs, line 11: I think in the other patterns we have this as numerical steps to show order of execution

patterns/bottlerocket/addons.tf
```diff
@@ -0,0 +1,130 @@
...
+  tags = local.tags
+}
+
+################################################################################
+# Karpenter resources
+################################################################################
+resource "helm_release" "karpenter_resources" {
+  name  = "karpenter-resources"
+  chart = "./karpenter-resources"
+  set {
```
bryantbiggs, line 104: optional nit - putting it in this form in examples is less digestible for readers. most readers will think and understand in terms of a `values.yaml` style, so if we can match that here, it makes it a little easier to grok whats going on. Something like:

```hcl
  values = [<<-EOT
    ec2nodeclass:
      securityGroupSelectorTerms:
        tags: ${module.eks.cluster_name}
      subnetSelectorTerms:
        tags: ${module.eks.cluster_name}
      ...
  EOT
  ]
```

patterns/bottlerocket/eks.tf
```diff
@@ -0,0 +1,143 @@
...
+
+    iam_role_attach_cni_policy = true
+  }
+
+  eks_managed_node_groups = {
+    bottlerocket = {
+      platform = "bottlerocket"
+
+      # The following specifies a custom ami_id, this can be provided by a data-source. Remove/comment to use the latest Bottlerocket AMI available.
+      ami_id = data.aws_ami.eks_bottlerocket.image_id
```
bryantbiggs, line 65: instead of this, can we just use `ami_release_version` and specify an older release version to start? that would keep everything "native" within managed nodegroups realm, and you won't need to inject the bootstrap user data, etc.

patterns/bottlerocket/eks.tf
```diff
@@ -0,0 +1,143 @@
...
+  eks_managed_node_group_defaults = {
+    ami_type       = "BOTTLEROCKET_x86_64"
+    instance_types = ["m5.large", "m5a.large"]
+
+    iam_role_attach_cni_policy = true
+  }
+
+  eks_managed_node_groups = {
+    bottlerocket = {
+      platform = "bottlerocket"
```
bryantbiggs, line 62: if AMI type is being set as default, then platform should be as well. or don't set the AMI type as default 

in the future, `platform` will be removed and we'll just simply rely on AMI type, but for now ... we have this sadness 😞

patterns/bottlerocket/eks.tf
```diff
@@ -0,0 +1,143 @@
...
+            kms_key_id            = module.ebs_kms_key.key_arn
+            delete_on_termination = true
+          }
+        }
+        xvdb = {
+          device_name = "/dev/xvdb"
+          ebs = {
+            encrypted             = true
+            kms_key_id            = module.ebs_kms_key.key_arn
+            delete_on_termination = true
```
bryantbiggs, line 87: future: we should show/talk about caching images on the data volume for faster pod startup. @stockholmux has more details on this

patterns/bottlerocket/main.tf
```diff
@@ -0,0 +1,48 @@
...
+data "aws_availability_zones" "available" {}
+
+data "aws_caller_identity" "current" {}
+
+locals {
+  name   = substr(basename(path.cwd), 0, 12)
+  region = "us-west-2"
+  tags = {
+    Sample     = local.name
+    GithubRepo = "github.com/aws-samples/amazon-eks-bottlerocket-mngnodegrp-terraform"
```
bryantbiggs, line 13: lets be consistent with the other patterns https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/1ce59fc329c9bfbd0bdb68d0bc5736a3fb4988ca/patterns/karpenter/main.tf#L50-L53

patterns/bottlerocket/vpc.tf
```diff
@@ -0,0 +1,39 @@
+################################################################################
+# VPC
+################################################################################
+locals {
+  azs = slice(data.aws_availability_zones.available.names, 0, 3)
+}
+
+variable "cidr_block" {
```
bryantbiggs, line 8: why do we need to make this a variable?

patterns/bottlerocket/versions.tf
```diff
@@ -0,0 +1,18 @@
+terraform {
```
bryantbiggs, line 1: I can't comment on the empty file, but just above we can remove the `variables.tf` - we don't have modules her so we don't enforce the standard module convention anymore

patterns/bottlerocket/eks.tf
rodrigobersa: I just removed the entire default section to fit other patterns.

bryantbiggs (approved):

PR #1896, opened 2024-03-01
refactor: Removing the existing Istio multi-cluster pattern to prepare for replacement with a refined multi-vpc pattern

vchintal:
# Description

Removal of the existing `istio-multi-cluster` pattern in preparation for a replacement with a refined multi-vpc pattern


bryantbiggs (commented):
we'll need to remove the doc page as well to avoid doc errors https://github.com/aws-ia/terraform-aws-eks-blueprints/blob/main/docs/patterns/istio-multi-cluster.md

bryantbiggs (approved):
thank you!

PR #1922, opened 2024-04-18
feat: Add pattern that demonstrates how to consume targeted ODCRs

bryantbiggs:
# Description

-  Add pattern that demonstrates how to consume targeted ODCRs

### Motivation and Context

- Provide example guidance on how to integrate ODCR consumption into EKS 


.github/scripts/mkdocs-hooks.py
```diff
@@ -0,0 +1,21 @@
+import os
```
bryantbiggs, line 1: this is how the supporting assets (i.e. - images) for a pattern are copied over into the built doc site. If you see the current Kubecost pattern, it works in markdown on GitHub, but not on the mkdoc site - I'll fix that after this PR using this code below

askulkarni2 (approved):
Thank for the PR! :shipit:

PR #1926, opened 2024-04-20
fix: Update EKS module min version to fix race condition during destroy

bryantbiggs:
# Description

- Update EKS module min version to fix race condition during destroy

### Motivation and Context

- Reference https://github.com/terraform-aws-modules/terraform-aws-eks/pull/3000


askulkarni2 (approved):
LGTM!

PR #1941, opened 2024-05-07
feat: Add pattern that demonstrates using ML capacity block reservation with self-managed node group

bryantbiggs:
# Description
- Add pattern that demonstrates using ML capacity block reservation with self-managed node group

### Motivation and Context
- Demonstrate how customers can use CBRs with EKS today; will update at later day to use EKS managed node group when that support is added to the service

### Additional Notes

In draft mode pending https://github.com/terraform-aws-modules/terraform-aws-eks/pull/3030


askulkarni2 (approved):
Awesome!

PR #1946, opened 2024-05-13
chore: Update targeted ODCR to show p5 to be reflect current usage

bryantbiggs:
# Description

- Update targeted ODCR to show p5 to be reflect current usage
	- Remove t3 capacity reservations and require reservation ARN be passed in via variable
	- Show usage with NVIDIA GPU instance and EFA which is the common use case today
- Add pod identity agent - start making it a standard on patterns
- Split out Helm charts to separate file to allow importing into docs site when necessary
- Add kubectl output per usual

### Motivation and Context

- While its great to have a pattern that can be deployed and see the reservations, etc. - it did not reflect the current target audience (p4/p5 users with EFA), which led to confusion. This updates the pattern to better reflect what the majority would be using today with capacity allocated in targeted ODCRs (tl;dr - meeting users where they are)

### DIFF
--- a/patterns/nvidia-gpu-efa/eks.tf
+++ b/patterns/nvidia-gpu-efa/eks.tf
@@ -15,9 +15,10 @@ module "eks" {
   cluster_endpoint_public_access           = true
 
   cluster_addons = {
-    coredns    = {}
-    kube-proxy = {}
-    vpc-cni    = {}
+    coredns                = {}
+    eks-pod-identity-agent = {}
+    kube-proxy             = {}
+    vpc-cni                = {}
   }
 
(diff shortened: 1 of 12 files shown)


patterns/nvidia-gpu-efa/eks.tf
```diff
@@ -71,18 +70,6 @@ module "eks" {
     default = {
       instance_types = ["m5.large"]
 
-      # Default AMI has only 8GB of storage
-      block_device_mappings = {
```
bryantbiggs, line 75: not strictly required for this architecture so removing to reduce "noise"

askulkarni2 (approved):
LGTM!

PR #1980, opened 2024-07-23
fix: ECR pull through cache pattern

rodrigobersa:
# Description

### Motivation and Context

Fit pattern to EKS Blueprints standards.

- Remove comments
- Default to local variables
- Set addons values
- Remove values.yaml files and variables.


patterns/ecr-pull-through-cache/addons.tf
```diff
@@ -1,3 +1,9 @@
+locals {
+  ecr_account_id = var.ecr_account_id != "" ? var.ecr_account_id : data.aws_caller_identity.current.account_id
```
bryantbiggs, line 2: lets remove account ID and region entirely - we can get this data from the provider/data sources

patterns/ecr-pull-through-cache/addons.tf
```diff
@@ -67,11 +109,28 @@ module "gatekeeper" {
...
+      preUninstall:
+        deleteWebhookConfigurations:
+          image:
+            repository: ${local.ecr_url}/docker-hub/openpolicyagent/gatekeeper-crds
+
+  EOT
+  ]
 
-  depends_on = [time_sleep.addons_wait_60_seconds]
+  depends_on = [module.eks_blueprints_addons]
```
bryantbiggs, line 135: why does it need to depend on? should we just add this into the addons module above under the generic `helm_releases`?

patterns/ecr-pull-through-cache/ecr.tf
```diff
@@ -1,19 +1,15 @@
...
+module "secrets_manager" {
   source  = "terraform-aws-modules/secrets-manager/aws"
-  version = "1.1.2"
+  version = "~> 1.1"
 
-  name          = "ecr-pullthroughcache/docker"
-  secret_string = jsonencode(var.docker_secret)
+  name                    = "ecr-pullthroughcache/docker"
+  secret_string           = jsonencode(var.docker_secret)
+  recovery_window_in_days = 0
```
bryantbiggs, line 7: we should add a note on why we are setting this to 0

patterns/ecr-pull-through-cache/variables.tf
```diff
@@ -1,61 +1,26 @@
...
 
-# defaults to local.region -> var.region
+# defaults to local.region
 variable "ecr_region" {
   type        = string
   description = "ECR repository AWS Region"
   default     = ""
 }
 
 variable "docker_secret" {
```
bryantbiggs, line 15: we should remove all variables except for those where we absolutely need users to supply information. Here, that looks like its just the `docker_secret` variable. We should also remove the defaults - we are requiring them to supply this, and by setting a default it will just fail silently

patterns/ecr-pull-through-cache/addons.tf
rodrigobersa: I kept those as different from the main region/account to illustrate that's possible to use an ECR on a different region/account, not required to be the same as the cluster deployment.

patterns/ecr-pull-through-cache/addons.tf
rodrigobersa: Fixing it, it should be  `depends_on = [module.eks.cluster_addons]`, since it requires VPC CNI to be up and running before start deploying, or it will break. 

I can also add on the -addons module.

bryantbiggs: because we are using `wait = true`? we should not wait and let the controllers reconcile themselves (or don't deploy these with Terraform 😉 )

patterns/ecr-pull-through-cache/main.tf
```diff
@@ -11,10 +8,13 @@ terraform {
     }
     helm = {
       source  = "hashicorp/helm"
-      version = ">= 2.13"
+      version = ">= 2.9"
```
candonov, line 11: why downgrading the helm version?

rodrigobersa: Not downgrading, it's just compatible with 2.9 or higher to match other patterns.

bryantbiggs (approved):
looks great, thank you!

PR #1995, opened 2024-08-12
fix: Availability Zone AZs Data Source to filter Local Zones

rodrigobersa:
# Description

Add a filter block on the availability zones data to exclude Local Zones.
Removing version filter for Karpenter helm chart on Bottlerocket pattern.

### Motivation and Context

Patterns deployments fail when Local Zones are enabled in the account. Adding filter to avoid the deployment on accounts with Local Zones enabled.
The parameter for the chart version is wrong on the Bottlerocket pattern, falling on the module default. Removing the version constraint.

### DIFF
--- a/patterns/blue-green-upgrade/modules/eks_cluster/main.tf
+++ b/patterns/blue-green-upgrade/modules/eks_cluster/main.tf
@@ -1,7 +1,7 @@
 # Required for public ECR where Karpenter artifacts are hosted
 provider "aws" {
   region = "us-east-1"
-  alias  = "virginia"
+  alias  = "ecr"
 }
 
 locals {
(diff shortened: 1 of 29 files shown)


patterns/blue-green-upgrade/modules/eks_cluster/main.tf
```diff
@@ -1,7 +1,7 @@
 # Required for public ECR where Karpenter artifacts are hosted
 provider "aws" {
   region = "us-east-1"
-  alias  = "virginia"
```
bryantbiggs, line 4: revert this change

patterns/bottlerocket/addons.tf
```diff
@@ -42,7 +42,6 @@ module "eks_blueprints_addons" {
   karpenter = {
     repository_username = data.aws_ecrpublic_authorization_token.token.user_name
     repository_password = data.aws_ecrpublic_authorization_token.token.password
-    version             = "v0.36"
```
bryantbiggs, line 45: is this not used?

PR #2062, opened 2025-01-16
docs: Fix vpc-cni-custoom-networking doc

ns-jsattler:
Just a simple syntax issue in the doc example.  Just missing a single }


bryantbiggs (approved):
thank you

PR #2065, opened 2025-01-23
refactor: Updates to privatelink-access based on latest working pattern

px-dsingh:
# Description
We have used the `privatelink-access` pattern internally and it mostly worked all good. In this PR included are some minor updates that we had to do to make it work:
- Bumping up module versions
- Simplifying security groups to use CIDRs
- Fixing health check to use TCP and not HTTPS as we are using an NLB
- Adding an egress security group rule to allow the NLB to communicate with the EKS cluster

### Motivation and Context
We needed to follow this pattern to connect two VPCs in different accounts and one was hosting an EKS cluster. Contributing back changes that we had to make in order to make the pattern work.

### Additional Notes
N/A

### DIFF
--- a/patterns/privatelink-access/client.tf
+++ b/patterns/privatelink-access/client.tf
@@ -11,7 +11,7 @@ locals {
 
 module "client_vpc" {
   source  = "terraform-aws-modules/vpc/aws"
-  version = "~> 5.0"
+  version = "~> 5.17"
 
   name = local.client_name
   cidr = local.vpc_cidr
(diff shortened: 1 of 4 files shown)


bryantbiggs:
Are you saying the current pattern is not working as intended per the README?

px-dsingh:
> Are you saying the current pattern is not working as intended per the README?

Correct yes, the targets didn't come up healthily for me behind the NLB as per the pattern.

bryantbiggs:
I've deployed the pattern as is this morning and its working as intended for me - could you try deploying it again, as is? 

Unfortunately, I don't think most of these changes are warranted or they are opening up more access than what is necessary

px-dsingh:
> I've deployed the pattern as is this morning and its working as intended for me - could you try deploying it again, as is?
> 
> Unfortunately, I don't think most of these changes are warranted or they are opening up more access than what is necessary

Ok no worries @bryantbiggs for me the health check and security groups were causing issues, but if it works for you then I'm happy to close this pull request.

PR #2072, opened 2025-01-30
feat: Add pattern for multi-node inference

bryantbiggs:
# Description

- Add pattern for multi-node inference

### Motivation and Context

- Several convos with users facing challenges trying to support multi-node inference

### DIFF
--- a/patterns/multi-node-vllm/ecr.tf
+++ b/patterns/multi-node-vllm/ecr.tf
@@ -8,9 +8,8 @@ module "ecr" {
 
   repository_name = local.name
 
-  create_lifecycle_policy         = false
-  repository_force_delete         = true
-  repository_image_tag_mutability = "MUTABLE"
+  create_lifecycle_policy = false
+  repository_force_delete = true
 
   tags = local.tags
 }
--- a/patterns/multi-node-vllm/ecr.tf
+++ b/patterns/multi-node-vllm/ecr.tf
@@ -1,0 +1,46 @@
+################################################################################
+# ECR Repository
+################################################################################
+
+module "ecr" {
+  source  = "terraform-aws-modules/ecr/aws"
+  version = "~> 1.6"
+
+  repository_name = local.name
+
+  create_lifecycle_policy         = false
+  repository_force_delete         = true
+  repository_image_tag_mutability = "MUTABLE"
+
+  tags = local.tags
+}
+
+################################################################################
+# Image Build & Push Script
+################################################################################
+
+data "aws_caller_identity" "current" {}
+
+resource "local_file" "vllm" {
+  content = <<-EOT
+  #!/usr/bin/env bash
+
+  aws ecr get-login-password --region ${local.region} | docker login --username AWS --password-stdin ${data.aws_caller_identity.current.account_id}.dkr.ecr.${local.region}.amazonaws.com
+
+  TAG=vllm-$(date +%Y%m%d_%H%M%S)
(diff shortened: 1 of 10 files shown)


patterns/multi-node-vllm/helm.tf
```diff
@@ -0,0 +1,78 @@
...
+}
+
+resource "helm_release" "kuberay_cluster" {
+  namespace        = "ray-cluster"
+  create_namespace = true
+  name             = "ray-cluster"
+  repository       = "https://ray-project.github.io/kuberay-helm/"
+  chart            = "ray-cluster"
+  version          = "1.2.2"
+}
```
askulkarni2, line 55: I don't think you need these. From what I can tell its managing the ray cluster internally.

bryantbiggs: let me try without - just to be clear, you're saying without both kuberay and ray-cluster, or just without ray-cluster?

patterns/multi-node-vllm/ecr.tf
```diff
@@ -0,0 +1,46 @@
...
+
+resource "local_file" "vllm" {
+  content = <<-EOT
+  #!/usr/bin/env bash
+
+  aws ecr get-login-password --region ${local.region} | docker login --username AWS --password-stdin ${data.aws_caller_identity.current.account_id}.dkr.ecr.${local.region}.amazonaws.com
+
+  TAG=vllm-$(date +%Y%m%d_%H%M%S)
+  docker build . --output type=image,name=${module.ecr.repository_url}:$${TAG},compression=zstd,force-compression=true,compression-level=9,oci-mediatypes=true,platform=linux/amd64
+  docker push ${module.ecr.repository_url}:$${TAG}
```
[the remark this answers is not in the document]
bryantbiggs: its uh ... not quick 😅 . let me add and with the hardware specs for reference

patterns/multi-node-vllm/helm.tf
askulkarni2: Yeah remove both. `kuberay-operator` CRUDs the CRDs invoked by `ray-cluster`, which is not happening with lws. Its doing a naive implementation of ray cluster via the `ray init` commands. Which also means it is not fault tolerant by default because Ray clusters should be using an external redis store for its state. We will need to dig a bit deeper on what happens if the LWS manager crashes because of an OOM or something.

patterns/multi-node-vllm/ecr.tf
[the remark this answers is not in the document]
bryantbiggs: added note to docs 
image

patterns/multi-node-vllm/eks.tf
```diff
@@ -0,0 +1,98 @@
...
+        {
+          content_type = "application/node.eks.aws"
+          content      = <<-EOT
+            ---
+            apiVersion: node.eks.aws/v1alpha1
+            kind: NodeConfig
+            spec:
+              instance:
+                localStorage:
+                  strategy: RAID0
```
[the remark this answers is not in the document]
bryantbiggs: added in https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/2072/commits/f4229bc7ee746578ced5edf39084fd47e9668067

patterns/multi-node-vllm/helm.tf
bryantbiggs: you were right! was able to remove both the operator and the cluster charts and everything still works as expected. updated in https://github.com/aws-ia/terraform-aws-eks-blueprints/pull/2072/commits/f4229bc7ee746578ced5edf39084fd47e9668067

askulkarni2 (approved):
LGTM!

PR #2082, opened 2025-03-06
fix: Correct multi-node vLLM Dockerfile dependencies

bryantbiggs:
# Description

- The previous implementation relied on a model that was small enough (7B) to fit into one instance (one L40S GPU), even though the pattern was intended to use multiple instances. This masked the underlying misconfiguration. I can only surmise that even when configured to use multi-node, if the model fits within the GPU(s) on a single node, it will do so automatically and without any warning (TBD). In order to ensure the pattern is configured correctly to use multi-node inference (and verify its configured correctly), a larger model (70B) is used to ensure it will not fit within a single L40S GPU found on g6e instances; this forces the pattern to be multi-node and fail if not
- The previous Dockerfile spec used the upstream vLLM pre-built image as a base from [here](https://github.com/vllm-project/vllm/blob/main/Dockerfile). However, that image doubles the NVIDIA components used and pulls in a number of libs/components that are not necessary, resulting in an image thats approaching 17GB. The updated image only pulls in whats necessary and is now 7.6GB unpacked on disk, and building in roughly half of the previous time required
- The LWS manifest has been updated for the larger model and number of nodes, as well as optimized parameters for running on g6e - but most importantly, now includes the EFA resources required to attach EFA devices to the pods which is ultimately the underlying issue in the previous (incorrect) configuration. This is now verified to be using EFA correctly as shown by the `via NET/Libfabric/0` in the NCCL debug logs (pro tip - don't just validate the inference endpoint with an inference request when using multi-node)

image

image

### Motivation and Context

- Correct the pattern to ensure its utilizing EFA and multi-node 


askulkarni2 (approved):
🔥

PR #2186, opened 2026-04-28
feat: Upgrade Python deprecated runtime for privatelink-access

vutny:
# Description

This PR updates old Python runtime for Lambda functions provided with `privatelink-access` module.
The upgrade path is from `python3.10` to `python3.13`.

### Motivation and Context
AWS deprecates Python 3.10 runtime starting from **Oct 31, 2026.**
More details are on official page: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-supported

- [x] Yes, I have tested the PR using my local account setup

### Additional Notes

Currently the latest available runtime is `python3.14`.
However, the module still uses old Terraform AWS Provider version `5.x`, which only allows to set `python3.13` as most recent.
The provider version at least `5.78` is required to set this runtime: https://github.com/hashicorp/terraform-provider-aws/blob/v5.78.0/CHANGELOG.md
Also, the `terraform-aws-modules/lambda/aws` module has been updated here to the last published version which still supports AWS Provider 5.x. That is `7.21.1` with the provider requirement of `5.79` at minimum. Upgrading lambda module further will cause broken provider version requirement, so it is fixed to that module version.

In addition, incorporated small fix from my previous PR #2153.

### DIFF
--- a/patterns/privatelink-access/lambdas/delete_eni.py
+++ b/patterns/privatelink-access/lambdas/delete_eni.py
@@ -42,6 +42,10 @@ def handler(event, context):
             unhealthyTargetIPAddress = targetHealthDescription["Target"]["Id"]
             unhealthyTargetIPAddresses.append(unhealthyTargetIPAddress)
 
+    if not unhealthyTargetIPAddresses:
+        logger.info("There are no unhealthy targets, quitting!")
+        return
+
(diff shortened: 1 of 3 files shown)


allamand (approved):
LGTM

